PR #17, opened 2017-04-05
Add postinstall-build to enable git:// install

### LINKED ISSUE #16: When adding "atom-languageclient" as dependency as stated in README.md, "apm install" will add a copy to node_modules/ with no "lib" or "build" which is hence unimportable
Warning: this might be due to a fundamental misunderstanding of node.js / npm on my side.

I tried the following things:
1. I made a new custom package closely following the README.md example code and the java language client atom package
2. I added "atom-languageclient" as dependency
3. I used `apm install` to install the resulting package

The following happened:
1. The package was copied into ~/.atom/packages/<package>
2. A subfolder was created `~/.atom/packages/<package>/node_modules`
3. There is now the dependency atom-languageclient inside `~/.atom/packages/<package>/node_modules/atom-languageclient`... but with no "build" folder which the main entry point in the package.json refers to, hence it's unimportable and my package doesn't actually work. Even more confusingly, the atom-languageclient version in that folder also lacks "lib" so running `npm run compile .` doesn't even allow me to actually build it.

What I expected instead is that `apm install` would also give me a working copy of `atom-languageclient` nested inside my custom package attempt so that things actually work.

hansonw:
Also this means that the original source can't be `.npmignore`d.
(This should be reverted once this is actually published).

Closes #16.

Taken from: https://github.com/npm/npm/issues/3055

Released under CC0.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,3 @@
-lib/
 .npmignore
-.babelrc
 .travis.yml
 appveyor.yml
--- a/package.json
+++ b/package.json
@@ -8,11 +8,13 @@
   "scripts": {
     "clean": "rm -rf build",
     "compile": "babel lib --out-dir build/lib && babel test --out-dir build/test",
+    "postinstall": "postinstall-build --only-as-dependency build \"npm run compile\"",
     "prepublish": "npm run clean && npm run compile",
     "test": "npm run compile && atom --test build/test",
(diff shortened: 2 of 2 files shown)


hansonw:
Argh one sec, this doesn't quite work

hansonw:
OK there we go. Needed the `.babelrc` too :)

PR #30, opened 2017-05-02
Highlight the word-at-buffer-position for hyperclick/definitions

### LINKED ISSUE #28: Definition provider should highlight only the word under the cursor
Currently, the definition/hyperclick providers query `textDocument/documentHighlight` to figure out what to highlight.

However, in both Nuclide and VSCode (https://github.com/Microsoft/vscode/blob/ff9f7b3baa6052863b55e8b6ef7c7c94d07beddf/src/vs/editor/contrib/goToDeclaration/browser/goToDeclaration.ts#L437), we simply highlight the word under the cursor (in Atom, we use the word regex as defined in the text editor).

This has the benefit of working for providers which don't implement word highlighting.
Also, it looks a bit weird to highlight all the references in the document.

@damieng would you be OK with this change/simplification?

hansonw:
Both Nuclide and VSCode use the word-at-buffer-position when cmd-clicking rather than using the `documentHighlight` LSP messages. This makes the Atom language client fall back to using the word when the `documentHighlight` capability isn't available.

The "_getRegexpRangeAtPosition" code is roughly borrowed from https://github.com/facebook/nuclide/blob/master/pkg/commons-node/range.js#L21.

See: https://github.com/Microsoft/vscode/blob/ff9f7b3baa6052863b55e8b6ef7c7c94d07beddf/src/vs/editor/contrib/goToDeclaration/browser/goToDeclaration.ts#L437

Closes #28. Contributed under CC0.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,0 +1,58 @@
+// @flow
+
+import {Range} from 'atom';
+
+export default class Utils {
+  /**
+   * Obtain the range of the word at the given editor position.
+   * Uses the non-word characters from the position's grammar scope.
+   */
+  static getWordAtPosition(editor: TextEditor, position: atom$Point): Range {
+    const scopeDescriptor = editor.scopeDescriptorForBufferPosition(position);
+    const nonWordCharacters = Utils.escapeRegExp(
+      editor.getNonWordCharacters(scopeDescriptor),
+    );
+    const range = Utils._getRegexpRangeAtPosition(
+      editor.getBuffer(),
+      position,
+      new RegExp(`^[\t ]*$|[^\\s${nonWordCharacters}]+`, 'g'),
+    );
+    if (range == null) {
+      return new Range(position, position);
+    }
+    return range;
+  }
+
+  static escapeRegExp(string: string): string {
+    // From atom/underscore-plus.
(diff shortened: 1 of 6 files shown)


lib/utils.js
```diff
@@ -0,0 +1,58 @@
...
+
+export default class Utils {
+  /**
+   * Obtain the range of the word at the given editor position.
+   * Uses the non-word characters from the position's grammar scope.
+   */
+  static getWordAtPosition(editor: TextEditor, position: atom$Point): Range {
+    const scopeDescriptor = editor.scopeDescriptorForBufferPosition(position);
+    const nonWordCharacters = Utils.escapeRegExp(
+      editor.getNonWordCharacters(scopeDescriptor.getScopesArray()),
```
Arcanemagus: This accepts a `ScopeDescriptor` directly, no need to pull out the array. (It _is_ non-public though.) (It [ends up](https://github.com/atom/atom/blob/v1.16.0/src/text-editor-registry.js#L399) being a simple [`atom.config.get`](https://atom.io/docs/api/v1.16.0/Config#instance-get).)

hansonw:
@Arcanemagus ah, that would appear to be the documented API (at least for `atom.config.get`). Thanks!

PR #33, opened 2017-05-03
Integrate document/hover with the Nuclide Datatip UI

### LINKED ISSUE #8: textDocument/hover
The [textDocument/hover](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#textDocument_hover) api allows additional information about a section of the document to be displayed in a hover/tool-tip style fashion.

## Challenges
- Nuclide has datatips for this but expect a react component not markdown as returned by LSP

hansonw:
This displays `document/hover` results by integrating with the Nuclide Datatip UI.
Note that this is stacked with #30 since this also uses the `wordAtPosition` utility - please take a look at that one first! :)

Screenshot using `flow-language-server`: (double scrollbar is caused by my having a plugged-in mouse; it's a bug we should fix)

screen shot 2017-05-03 at 1 20 31 pm

Closes #8.

Released under CC0.

### DIFF
--- a/lib/adapters/nuclide-datatip-adapter.js
+++ b/lib/adapters/nuclide-datatip-adapter.js
@@ -1,0 +1,65 @@
+// @flow
+
+import {
+  LanguageClientConnection,
+  type MarkedString,
+  type ServerCapabilities,
+} from '../languageclient';
+import Convert from '../convert';
+import Utils from '../utils';
+
+export default class NuclideDatatipAdapter {
+  static canAdapt(serverCapabilities: ServerCapabilities): boolean {
+    return serverCapabilities.hoverProvider === true;
+  }
+
+  async getDatatip(
+    connection: LanguageClientConnection,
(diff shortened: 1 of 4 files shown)


lib/adapters/nuclide-datatip-adapter.js
```diff
@@ -0,0 +1,61 @@
...
+    point: atom$Point,
+  ): Promise<?nuclide$Datatip> {
+    const documentPositionParams = Convert.editorToTextDocumentPositionParams(
+      editor,
+      point,
+    );
+
+    const hover = await connection.hover(documentPositionParams);
+    // This intentionally covers both empty strings and empty arrays.
+    if (hover == null || hover.contents.length === 0) {
```
damieng: Do we also want to check hover.contents isn't null?

hansonw: Good point, I forgot everything in Typescript is nullable by default D:

PR #39, opened 2017-05-21
Add optional ipc transport

josa42:
The [VSCode Language Server - Client](https://github.com/Microsoft/vscode-languageserver-node/tree/master/client) has the option to use [ipc](https://github.com/Microsoft/vscode-languageserver-node/blob/master/client/src/main.ts#L230) besides stdio message transports.

I am currently working on a Atom adaption of the [Ember Language Server](https://github.com/emberwatch/ember-language-server) which is using this.

### DIFF
--- a/lib/auto-languageclient.js
+++ b/lib/auto-languageclient.js
@@ -88,6 +88,11 @@ export default class AutoLanguageClient {
   postInitialization(InitializationResult: ls.InitializeResult): void {
   }
 
+  // Determine whether to use ipc or stdio transport
+  useIpc() {
+    return false;
+  }
+
   // Default implementation of the rest of the AutoLanguageClient
   // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


damieng:
I wonder if we should go with a general "Connection Type" style return that returns 'stdio', 'process' or 'socket' so that any future options don't end up with a confusing set.

josa42:
That is a good point. I did not want to break the current behavior.

I would introduce an enum `ConnectionType`:

```
type ConnectionType = "stdio" | "socket" | "ipc"
```

and `getConnectionType()` in `AutoLanguageClient` with this default implementation:

```
getConnectionType() : ConnectionType {
  const hasSocket = this.socket != null;
  return hasSocket ? "socket" : "stdio"
}
```

Would you agree with that?

lib/auto-languageclient.js
```diff
@@ -137,9 +145,19 @@ export default class AutoLanguageClient {
   }
 
   createRpcConnection(process: child_process$ChildProcess): rpc.Connection {
-    const hasSocket = this.socket != null;
-    const reader = hasSocket ? new rpc.SocketMessageReader(this.socket) : new rpc.StreamMessageReader(process.stdout);
-    const writer = hasSocket ? new rpc.SocketMessageWriter(this.socket) : new rpc.StreamMessageWriter(process.stdin);
+    let reader, writer;
+    const connectionTyp = this.getConnectionType();
```
damieng: missing an e on connectionType ?

damieng:
If you want to fix the typo on connectionTyp I'll get this PR merged :)

lib/auto-languageclient.js
josa42: 🙃 Oops

josa42:
Typo is fixed. Sorry about that.

damieng:
Thanks for contributing!

josa42:
Thank you.

PR #40, opened 2017-05-23
Add nuclide code format support

### LINKED ISSUE #27: Use the Nuclide code formatting API
I reworked it to work with TextEdits. I think this should be sufficient for the LSP range formatting API (the textedit type doesn't match exactly but it's really close.)

API: https://github.com/facebook/nuclide/blob/master/pkg/nuclide-code-format/lib/types.js#L24
Nuclide TextEdit: https://github.com/facebook/nuclide/blob/master/pkg/nuclide-textedit/lib/rpc-types.js

I'll probably send in a PR for this when I get the time.

damieng:
Fixes #27 

@hansonw I'm surprised by the behavior of Nuclide's "Code Format" command without any range selected.  Based on https://github.com/facebook/nuclide/blob/master/pkg/nuclide-code-format/lib/CodeFormatManager.js#L114 I expected to see no selection become whole file but I am not getting that in my callback - range is just set to the cursor position where end and start are both the same - effectively no range.

### DIFF
--- a/flow-libs/nuclide-providers.js.flow
+++ b/flow-libs/nuclide-providers.js.flow
@@ -189,3 +189,10 @@ export interface nuclide$CodeFormatProvider {
   selector: string,
   inclusionPriority: number,
 }
+
+export type nuclide$TextEdit = {
+  oldRange: atom$Range,
+  newText: string,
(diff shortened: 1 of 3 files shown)


hansonw:
Hmm that's odd indeed. Will try to repro..

PR #49, opened 2017-06-16
Minor tweaks to diagnostics/find-references adapters

hansonw:
In diagnostics: Use `linterName` for the source, rather than the description. This requires https://github.com/steelbrain/linter/pull/1493 but until then it'll be ignored in favor of the language server's name (which is fine).

In find-references: when there's no references, make sure an empty array is returned rather than `null`. The UI uses this to distinguish between "no references" and "can't find references".

Released under CC0.


damieng:
Awesome!

PR #53, opened 2017-06-28
Export LinterPushV2Adapter and let override diagnosticToV2Message

janos-ss:
Export `LinterPushV2Adapter` and make it possible to override `LinterPushV2Adapter.diagnosticToV2Message` in derived classes, in order to customize the messages sent to linter.

### DIFF
--- a/lib/auto-languageclient.js
+++ b/lib/auto-languageclient.js
@@ -59,6 +59,10 @@ export default class AutoLanguageClient {
     throw Error('Must override startServerProcess to start language server process when extending AutoLanguageClient');
   }
 
+  newLinterPushV2Adapter(connection: ls.LanguageClientConnection): LinterPushV2Adapter {
+    return new LinterPushV2Adapter(connection);
+  }
+
   // You might want to override these for different behavior
   // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


janos-ss:
My main need is to customize the messages in my [upcoming linter](https://github.com/janos-ss/sonarlint-atom/blob/master/sonarlint/lib/sonarlint.js), namely, set `url` and `solutions` fields.

If this change needs more work, please let me know, I'd really love to get this merged (and published) soon. Thanks!

damieng:
Thanks for this PR. The plan is indeed to expose the adapters so people can mix and match as they need.  

I'll take a look at this - I suspect I've just caused some merge conflicts with some clean-up work.

janos-ss:
Glad you like it. I resolved the conflicts.

Btw, if you decide to merge my PRs, always feel free to squash them into one commit to keep the history nice and flat, no need to keep the individual steps.

lib/auto-languageclient.js
```diff
@@ -59,6 +59,10 @@ export default class AutoLanguageClient {
     throw Error('Must override startServerProcess to start language server process when extending AutoLanguageClient');
   }
 
+  newLinterPushV2Adapter(connection: ls.LanguageClientConnection): LinterPushV2Adapter {
```
damieng: Given how many adapters we have and how many more to come I'm wondering (out loud) if we might be better off just having a single method do them all or some simple IoC/Dependency Injection pattern.

janos-ss: A general IoC/DI pattern sounds nice, but isn't it overkill for now? (I try to avoid speculative over-engineering.)

damieng:
i think i'll tweak the construction pattern but will start by getting this merged. Thanks!

henryju:
Seems the ability to replace the LinterPushV2Adapter was lost in recent versions. I want to set the url attribute on messages.

PR #54, opened 2017-06-28
Adapt DocumentSyncAdapter to LSPv3

janos-ss:


janos-ss:
:shit: This is not enough. Diagnostics are correctly received on file open, but on on-the-fly. Investigating now. Closing until fixed.

damieng:
Thanks for this.

I wasn't planning on adding V3 support initially but this looks like V3 servers might be backward incompatible without this?

janos-ss:
> I wasn't planning on adding V3 support initially but this looks like V3 servers might be backward incompatible without this?

No, the incompatibility for this one is at the client side: the type of `ServerCapabilities.textDocumentSync`in v2 is `number`, but in v3 it's `TextDocumentSyncOptions | number`. So while a v2 server sends for example `{ textDocumentSync: TextDocumentSyncKind.Full }`, a v3 server may send `{ textDocumentSync: { change: TextDocumentSyncKind.Full } }`. And currently the client cannot handle this latter case, because `{ change: TextDocumentSyncKind.Full } != TextDocumentSyncKind.Full`

My change fixes the incompatibility of the client by checking if the value received from the server is an `object` *and* has a `change` property, then use that. Otherwise fall back to using the value directly.

I'm a bit concerned that for proper support of v3, the full spectrum of changes from v2 -> v3 should be reviewed, otherwise I'm not sure if there might be hidden issues. For example, my upcoming [linter package](https://github.com/SonarSource/sonarlint-atom/releases/tag/0.0.0-M5) seems to work fine with the [temporary fork of atom-languageclient](https://github.com/SonarSource/atom-languageclient) (incorporating all my PRs), but there's a nasty bug somewhere I've been struggling to fix (language server stops receiving messages after 100~ keystrokes), and maybe the problem comes from my changes.

janos-ss:
The problem in my [upcoming linter package](https://github.com/SonarSource/sonarlint-atom/releases/tag/1.0-RC1) is now fixed, and it was not related to the language protocol. (I had to consume the `stdout` of the spawned language server process.)

So I can say that with the change in this PR to adapt to the changed server capability message coming from v3, the integration with language server seems to work fine.

wz1000:
Any update on when this would be merged?

PR #57, opened 2017-07-07
Sync documents for supported languages only

### LINKED ISSUE #56: Document synchronization actions are triggered for unsupported languages
The language client gets activated when working with files whose language matches `getGrammarScopes`. But once activated, document synchronization actions are triggered for all files, because the editor selector function passed to `DocumentSyncAdapter` disregards `getGrammarScopes`.

janos-ss:
Fixes #56

### DIFF
--- a/lib/auto-languageclient.js
+++ b/lib/auto-languageclient.js
@@ -176,7 +176,7 @@ export default class AutoLanguageClient {
     if (DocumentSyncAdapter.canAdapt(server.capabilities)) {
       server.disposable.add(
         new DocumentSyncAdapter(server.connection, server.capabilities.textDocumentSync,
-          this.shouldSyncForEditor(editor, server.projectPath)));
+          editor => this.shouldSyncForEditor(editor, server.projectPath)));
     }
 
     server.linterPushV2 = new LinterPushV2Adapter(server.connection);
@@ -185,8 +185,8 @@ export default class AutoLanguageClient {
     }
   }
 
-  shouldSyncForEditor(editor: atom$TextEditor, projectPath: string): atom$TextEditor => boolean {
-    return editor => this._isFileInProject(editor, projectPath) && this._shouldStartForEditor(editor);
+  shouldSyncForEditor(editor: atom$TextEditor, projectPath: string): boolean {
+    return this._isFileInProject(editor, projectPath) && this._shouldStartForEditor(editor);
   }
 
   _isFileInProject(editor: atom$TextEditor, projectPath: string): boolean {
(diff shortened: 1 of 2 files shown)


lib/auto-languageclient.js
```diff
@@ -171,6 +171,18 @@ export default class AutoLanguageClient {
...
 
+  newEditorSelector(editor: atom$TextEditor, basepath: string): atom$TextEditor => boolean {
+    return editor => this._isFileInProject(editor, basepath) && this._isSupportedLanguage(editor);
+  }
+
+  _isFileInProject(editor: atom$TextEditor, basepath: string): boolean {
+    return (editor.getURI() || '').startsWith(basepath);
+  }
+
+  _isSupportedLanguage(editor: atom$TextEditor): boolean {
```
damieng: We should then call this from `shouldStartForEditor`

lib/auto-languageclient.js
```diff
@@ -171,6 +171,18 @@ export default class AutoLanguageClient {
     }
   }
 
+  newEditorSelector(editor: atom$TextEditor, basepath: string): atom$TextEditor => boolean {
```
damieng: Let's rename this `shouldSyncForEditor` and change the second arg to `projectPath` for consistency.

lib/auto-languageclient.js
```diff
@@ -171,6 +171,18 @@ export default class AutoLanguageClient {
...
+    return editor => this._isFileInProject(editor, basepath) && this._isSupportedLanguage(editor);
+  }
+
+  _isFileInProject(editor: atom$TextEditor, basepath: string): boolean {
```
damieng: Ditto here with the second arg being `projectPath`

janos-ss:
I did the requested changes.

I have a question related to this. Why do we have the `_isFileInProject` condition at all? Why not include all open files? Some of my testers found it unexpected to not see issues in files that they just dragged into the Atom window, and so were not part of the project. What would be the harm in including those?

damieng:
When you start a language server you have to give it a 'root path' which is effectively the top folder where project configuration files live.

Subsequent to that they only expect to receive events relating to files beneath that path.

We do currently support if you have more than one folder correctly starting two projects and ensuring the right server gets the right file events.

We don't yet however support ad-hoc opening of a single file. To do that correctly we'd need to identify it wasn't in an existing path and then traverse up the file system folders until we find what we think would likely be the root then start the server with that.  Deciding what constitutes a project root would be language specific (e.g. `.flow` for Flow, `.sln` or `.csproj` for C# etc)

janos-ss:
Thanks for the explanation.

Btw I don't understand why `continuous-integration/appveyor/pr` fails. I rebased on `master` and all tests pass for me.

PR #69, opened 2017-07-21
Set descriptionMarkdown property for Atom auto-complete suggestion

BoykoAlex:
AutoComplete+ suggestion model has `descriptionMarkdown` property for Markdown format proposal documentation. (Think it's safe for plain text as well)
Although LS protocol doesn't specify that `documentation` is `MarkedString` VSCode language client accepts Markdown format documentation, so does Eclipse client. Looks like it's easy to add support for this in Atom too.

### DIFF
--- a/lib/adapters/autocomplete-adapter.js
+++ b/lib/adapters/autocomplete-adapter.js
@@ -80,6 +80,7 @@ export default class AutocompleteAdapter {
   //
   // Returns an AutoComplete+ suggestion.
   static basicCompletionItemToSuggestion(item: CompletionItem): atom$AutocompleteSuggestion {
+    //noinspection JSAnnotator
     return {
       text: item.insertText || item.label,
       displayText: item.label,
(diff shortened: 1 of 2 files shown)


damieng:
I wonder if we should have some detection logic that tries to figure out if it's markdown or not and then sets the appropriate field.

I'm assuming right now that with that change it always interprets this as markdown because that property is always set.

BoykoAlex:
Plain text can be considered as markdown and markdown can be considered as plain text. Think the check that we could put in could only rely on some likelyhood of text being Markdown... Perhaps we could expect `[comment]: <> (Markdown)` a markdown comment at start of the doc string to interpret it as markdown... What do you think?

BoykoAlex:
Think it's best to wait for the protocol to have something that reliably tells us whether `documentation` is Markdown or anything else. (https://github.com/Microsoft/language-server-protocol/issues/111)
Don't think it'll be such a bad thing if plain text will be rendered as markdown though, but there might be regressions.

damieng:
I think you're right - until the protocol allows specification we shouldn't use both. 

There's a chance ASCII will be misinterpreted if we make this the default. We either need a config switch or an easy code-path to change the behavior.

damieng:
Going to merge this and then change it to be parameter driven so package authors can set a flag if they want markdown formatted descriptions. Thanks!

Issue #76, opened 2017-07-26
Overriding default grammar for file type doesn't start LS


BoykoAlex:
Suppose i have a LS that start for grammar `source.my-lang` and the grammar applies to `.my-lang` files.
I open a file `blah.txt`. It opens with `Plain Text` then switch the grammar to `my-lang` and expect the LS to start and services be provided. Also when i switch back to `Plain Text` services should be disconnected from this file as well as errors/warning cleared, LS shut down if not used.

Same goes if you try to override grammar for the grammar LS is running for. The LS won't shut down and then start once grammar is switched back.

Looks like each opened editor should be observed for grammar changes. Working on the PR for this.

BoykoAlex:
PR: https://github.com/atom/atom-languageclient/pull/77

PR #77, opened 2017-07-27
Fixes around editor grammar override feature. Tracking grammar changes for LS start/stop and file syncing

BoykoAlex:
Attempt to fix https://github.com/atom/atom-languageclient/issues/76

One issue left is to wipe out errors/warnings (markers) when switching from LS backed grammar to non LS grammar. Not sure how to get it fixed, need a hint

### DIFF
--- a/lib/adapters/document-sync-adapter.js
+++ b/lib/adapters/document-sync-adapter.js
@@ -73,16 +73,44 @@ export default class DocumentSyncAdapter {
   //
   // * `editor` A {TextEditor} to consider for observation.
   observeTextEditor(editor: atom$TextEditor): void {
+    const listener = editor.observeGrammar(grammar => this._handleGrammarChange(editor));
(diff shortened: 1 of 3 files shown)


damieng:
I think what we need to do to clear out those errors is send a file close event in that scenario.

BoykoAlex:
Hmm... think I'm already doing that...I have a call `syncAdapter.didDestroy();` in `server-manager.js` line 83.
I see that `textDocument/didClose` is message is sent. The next message i see is `textDocument/publishDiagnostics` with empty diagnostics array... 
I have also tried manually setting empty diagnostics array by calling `linterAdapter#captureDiagnostics(...)` and that doesn't work either... Seems like the DOM isn't updated, but model is...

damieng:
I want to rework some of the tracking code so I'll merge your work in and then take care of the issue you pointed out. Thanks for contributing!

PR #78, opened 2017-08-03
Upgrade to Flow 0.52 and run Flow on Travis

hansonw:
Note that this caught a very real bug: `getActiveServers()` would empty the array! (slice vs splice)
The rest are pretty trivial changes. I'll add Flow to the Travis CI config as well.

Contributed under CC0.

### DIFF
--- a/script/cibuild
+++ b/script/cibuild
@@ -1,5 +1,8 @@
 #!/bin/sh
 
+# Exit if any single command fails.
+set -e
+
 echo "Downloading latest Atom release..."
 ATOM_CHANNEL="${ATOM_CHANNEL:=stable}"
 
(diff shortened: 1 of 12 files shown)


hansonw:
It's passing :) https://travis-ci.org/atom/atom-languageclient/builds/260833314

script/cibuild
```diff
@@ -94,10 +97,8 @@ if [ -n "${APM_TEST_PACKAGES}" ]; then
   done
 fi
 
+echo "Running flow..."
+node_modules/.bin/flow
```
damieng: I wonder if we should put this in the standard package.json for build or lint?

hansonw:
OK, I added:

- `npm run lint` -> runs Flow (and hopefully ESLint once that's clean)
- `npm test` -> now runs `npm run lint` too
- Travis just runs `npm test` now

PR #87, opened 2017-08-10
[RFC] Lint clean and enable Prettier

hansonw:
This enables https://github.com/prettier/prettier-eslint and https://github.com/prettier/eslint-config-prettier to ensure that all code under `lib/` and `test`/ is lint clean and auto-formatted!

If you're using `atom-ide-diagnostics` with `linter-eslint`, you can hit alt-shift-A to automatically format your code to match Prettier :)

I mostly just ran `eslint --fix lib test`, but there were some manual changes necessary. I separated them into separate commits.

Finally, this enables linting on `npm test`, `npm run lint`, and Travis so hopefully it'll stay clean :)

Released under CC0.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,24 @@
 module.exports = {
-    "extends": "fbjs-opensource"
+  'extends': [
+    'fbjs-opensource',
+    'prettier',
+  ],
+  'plugins': [
+    'flowtype',
+    'prettier',
+  ],
+  'rules': {
+    'babel/no-await-in-loop': 0,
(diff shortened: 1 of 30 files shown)


hansonw:
Updated to use a width of 120 to be more consistent with the current style.
There's still a few function definitions that had to be split.. but I think that's a good thing overall?

hansonw:
Once you code with automatic formatting, you can't go back! (If you have `linter-eslint` installed, you can hit `alt-shift-A` in atom-ide to format the file).

PR #111, opened 2017-09-22
Autocomplete: Take priority over the default provider

lgeiger:
The autocompletions provided by the language server are usually more relevant than the default provider.

This PR increases the priority in order to display the completions above the default provider.

### master
master

### PR 
PR

Original issue: https://github.com/lgeiger/ide-python/issues/16#issuecomment-331579161


damieng:
Thanks!

Issue #114, opened 2017-09-24
Support CompletionItem additionalTextEdits and command


nwolverson:
`additionalTextEdits` can be used to do something like inserting import statements on completion, and `command` can be used for this and much more. These can be implemented with autocomplete+ method `onDidInsertSuggestion`.

PR #115, opened 2017-09-24
Add support for CompletionItem.command

nwolverson:
Maybe controversial smuggling extra data around with the completion item, which we get returned to us in `onDidInsertSuggestion`. Maybe there's a better way, also I have never used flow so this is probably not ideal.

I think the same strategy as for `command` can be used for additional text edits - i.e. on the didInsert callback the additional edits can be retrieved and applied to the document.

#114

### DIFF
--- a/flow-libs/atom.js.flow
+++ b/flow-libs/atom.js.flow
@@ -1758,11 +1758,18 @@ type atom$AutocompleteRequest = {
   activatedManually?: boolean,
 };
 
+type atom$AutocompleteDidInsert = {
+  editor: TextEditor,
+  triggerPosition: atom$Point,
+  suggestion: atom$AutocompleteSuggestion
+}
+
 type atom$AutocompleteProvider = {
   selector: string,
   getSuggestions: (
     request: atom$AutocompleteRequest,
   ) => Promise<?Array<atom$AutocompleteSuggestion>>,
+  onDidInsertSuggestion?: (arg: atom$AutocompleteDidInsert) => void,
   disableForSelector?: string,
   inclusionPriority?: number,
   excludeLowerPriority?: boolean,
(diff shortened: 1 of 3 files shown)


damieng:
I like the idea of providers being able to provide their own OnDidInsertSuggestion that gets returned as part of it and we could include a command binder - I think though that the two should be up to the ide-x author to wire up unless there's a precedence somewhere?

nwolverson:
My thinking was that a CodeAction specifies a `command` which is wired up to execute a command on the server automatically https://github.com/atom/atom-languageclient/blob/5e1d6641d2e53b6334208c1efed5be2e94412d21/lib/adapters/code-action-adapter.js#L60 and similarly CompletionItem has a `command`.

But certainly (in both cases) I could see that might not be what you want. I think eg in VSCode the command is automatically executed on the client side but those commands are then forwarded to the server.

I'm not sure I follow what you are suggesting particularly "command binder" - there is some `AutoLanguageClient.onDidInsertSuggestion` which can be overridden and gets given the `command` (or even the original CompletionItem)?

nwolverson:
@damieng is this closer to what you were suggesting? Expose `onDidInsertSuggestion` to be overridden, passing the original completion item entirely, as well as the autocomplete+`onDidInsertSuggestion` argument, making it easy to wire up command etc.

damieng:
Here's my revised idea - basically just implementing the onDidInsertSuggestion hook - used in conjunction with the new conversion hook api this should give you the flexibility to store and re-obtain the original completion item etc.  

Will merge today unless there are objections.

nwolverson:
Sorry I wasn't able to get back to this, thanks for making those changes - looks good to me.

damieng:
No problem - glad we found a solution that works.

PR #117, opened 2017-09-29
Flow types for busy signal service

damieng:


hansonw:
nice :) minor comment: might be good to make sure the busy message gets disposed when the whole thing gets disposed too

damieng:
I also need to track each server path separately to ensure the shutdown messages don't dispose each other.

damieng:
Changed this to only be the flow types exposed by the busy signal service.

Issue #125, opened 2017-10-23
didChangeWatchedFiles never-ending recursion when a server process writes build artifacts to a subdirectory


parasyte:
This was originally reported against `ide-rust`: https://github.com/mehcode/atom-ide-rust/issues/9 The language server `rls` writes build artifacts to a subdirectory of the atom project. `atom-languageclient` has a FS watch on the project directory to trigger the client's language server. In this particular case, the language server is triggers the FS watch, which causes an infinite loop of file changes.

The easiest solution I can imagine here is to ignore all changes in file paths matching `.gitignore`. It would solve this specific manifestation of the bug, and provide an escape hatch for similar clients if they hit the problem in the future.

FWIW there is a simple workaround patch provided in [this comment](https://github.com/mehcode/atom-ide-rust/issues/9#issuecomment-338647456) which hardcodes the `rls` artifact directory.

damieng:
We could add an ignore for .gitignore entries but that would involve taking a dependency on a wildcard matching library and filtering for every single entry.

Really language servers themselves should not be treating didChangeWatchedFiles as an instruction to rebuild all when it is intended just to be a file notification.

alexheretic:
It could be nice to allow clients to implement their own filtering in addition to any server filtering. This could be done with a extendible method on `AutoLanguageClient`.

```javascript
class AutoLanguageClient {
  /**
   * `didChangeWatchedFiles` message filtering, override for custom logic.
   * @param {string} filePath path of a file that has changed in the project path
   * @return {boolean} false => message will not be sent to the server
   */
  filterChangeWatchedFiles(filePath) {
    return true;
  }
  ...
}
```
That function can be overridden by clients providing client specific filtering, and passed to (and called within) `ServerManager` instances.

```javascript
class ServerManager {
  projectFilesChanged(fileEvents) {
    ...
      for (const fileEvent of fileEvents) {
        if (this._changeWatchedFileFilter(fileEvent.path) === false) {
          continue
        }
    ...
  }
  ...
}
```
This way clients not interested in filtering are unaffected. I can raise a PR if this seems workable?

Arcanemagus:
It would be much better if the language servers handled this themselves, after all the _entire_ point of them is to consolidate the work of implementing things like that in a single place for everyone that uses them.

alexheretic:
[...]

PR #127, opened 2017-10-27
Allow client didChangeWatchedFiles filtering

alexheretic:
This change will allow clients to help address #125. Even when ignored server side this can produce a lot of message churn when the language server is spamming files somewhere in the project, ie rls.

Here's a screen just to show you the possible magnitude. This is _with_ my patch having the rls ignore these btw.
debug-no-filter

With this patch and *ide-rust* popping a filter into the equation the situation is improved.
```js
// ide-rust
class RustLanguageClient extends AutoLanguageClient {
  ...
  filterChangeWatchedFiles(filePath) {
    return !filePath.includes('/.git/') && !filePath.includes('/target/rls/')
  }
}
```

screenshot from 2017-10-27 19-08-49

### DIFF
--- a/lib/auto-languageclient.js
+++ b/lib/auto-languageclient.js
@@ -106,7 +106,12 @@ export default class AutoLanguageClient {
   activate(): void {
     this.name = `${this.getLanguageName()} (${this.getServerName()})`;
     this.logger = this.getLogger();
-    this._serverManager = new ServerManager(p => this.startServer(p), this.logger, e => this.shouldStartForEditor(e));
+    this._serverManager = new ServerManager(
+      p => this.startServer(p),
+      this.logger,
+      e => this.shouldStartForEditor(e),
+      path => this.filterChangeWatchedFiles(path),
+    );
     this._serverManager.startListening();
     process.on('exit', () => this.exitCleanup.bind(this));
   }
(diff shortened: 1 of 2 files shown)


lib/server-manager.js
```diff
@@ -228,6 +231,10 @@ export class ServerManager {
     for (const activeServer of this._activeServers) {
       const changes = [];
       for (const fileEvent of fileEvents) {
+        if (this._changeWatchedFileFilter(fileEvent.path) === false) {
+          continue;
+        }
+
         if (fileEvent.path.startsWith(activeServer.projectPath)) {
```
damieng: Let's remove this if condition and make it the default.

lib/auto-languageclient.js
```diff
@@ -438,4 +443,13 @@ export default class AutoLanguageClient {
...
     });
   }
+
+  /**
+   * `didChangeWatchedFiles` message filtering, override for custom logic.
+   * @param filePath path of a file that has changed in the project path
+   * @return false => message will not be sent to the language server
+   */
+  filterChangeWatchedFiles(filePath: string): boolean {
+    return true;
```
damieng: If we make this filterChangeWatchedFiles method take the filePath and the projectPath we can move the default here while also providing more flexibility.

lib/server-manager.js
damieng: Actually... I think I'm okay with this as-is.

alexheretic:
Nice one thanks for reviewing!

Issue #89, opened 2017-08-15
Provide better support for triggering AutoComplete


damieng:
Right now we try and call textDocument/complete for every single keystroke as that is how autocomplete-plus works.

This is too expensive for many language servers (e.g. ide-typescript) and instead we need some idea of when to trigger and the ability to filter client-side in memory, e.g.

1. Request textDocument/complete when '.' is typed or ctl-space (manual trigger)
2. If the results were isComplete true then client-side filter using fuzzaldrin for further typing
3. If the results were isComplete false then call textDocument/complete again with further typing
4. Correctly handle backspace in scenarios 2-3

A version of this without isComplete handling is currently hacked into ide-typescript at https://github.com/atom/ide-typescript/blob/master/lib/main.js#L19

hansonw:
We have a version of this exact idea over at https://github.com/facebook/nuclide/blob/master/pkg/commons-atom/AutocompleteCacher.js, if you need to steal anything :P

rcjsuen:
Seems to me like `textDocument/complete` requests should only be sent when manually triggered or when one of the `triggerCharacters` as defined by `serverCapabilities.completionProvider` are hit.

damieng:
Yes, that is correct. Have worked around this in the typescript provider but need to build this into the language client library directly.

deankevorkian:
In the current state, with the hack in ide-typescript, the client-side filter is done by the ide-* package, and has full control on updating the `server.currentSuggestions`. 
Since we want to move things to the atom language client, who's going to be in charge of editing the currentSuggestions? 
since the fuzzy filtering will be done in the languageclient package, and there's no logic for the ide-* to take part in,  the suggestions should be modified right after the filtering is done, by the languageclient itself. 
On the other hand, if we're calling `textDocument/complete`, we don't want the languageclient to modify server.currentSuggestions and instead, only resolve the returned suggestions from the LSP client - because the ide-typescript package, for example, might want to control it - maybe ranking suggestions by some config parameters (show variables before methods) etc.

To avoid inconsistency I think the languageclient should never modify the currentSuggestions, only resolve suggestions for the ide-* package. But this means the ide-* will **always** have to modify the `server.currentSuggestions` - even if no results were returned. does this make sense?

damieng:
Checked in.

PR #128, opened 2017-10-28
(WIP) Provide better support for triggering AutoComplete

deankevorkian:
Refs #89 . 
In this PR I only try to handle scenarios 1,2 and 3. (Still not handling backspace correctly in scenarios 2 and 3).
I was thinking that the following PR could be a decent beginning of providing better support for the triggering AutoComplete (IMHO).

I changed the way the `AutoCompleteAdapter`'s `getSuggestions` returns suggestions from the `LanguageClient`. Instead of simply returning a list of the suggestions, straightly in an Array of autocomplete suggestion objects, it returns a 'Response' object that stores 2 properties - the actual suggestions (Atom AutoComplete suggestion objects, of course) and an "isComplete" indication that indicates if the results were complete or not (from the "IsIncomplete" indication from the LSP specification).

If the LS returns a `CompletionList` object, I can use its "IsIncomplete" indication. But if it doesn't, and returns a list of `CompletionItem` instead, I just store `IsComplete: true` as a default choice (maybe it should be false to force re-calculation?)

I also changed the AutoComplete adapter's tests to support the new structure returned from the `getSuggestions` method.

There's still work to do (handle backspaces and ensure we reset `areCurrentSuggestionsComplete` when we should), just wanted to know if this looks like a right direction to go with.. this means we have to break the `AutoCompleteAdapter`'s contract, but the `AutoLanguageClient`'s `getSuggestions~ still remains with the same contract.

@damieng, thoughts? should I keep going with this direction?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,6 @@
 module.exports = {
   'extends': [
     'fbjs-opensource',
-    'prettier',
   ],
   'plugins': [
     'flowtype',
@@ -11,6 +10,7 @@ module.exports = {
     'camelcase': 0,
     'flowtype/no-weak-types': 0,
     'flowtype/object-type-delimiter': 0,
+    'max-len': [2, 160],
     'no-new': 0,
     'linebreak-style': 0,
   },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,7 +5,6 @@ module.exports = {
   ],
   'plugins': [
     'flowtype',
-    'prettier',
   ],
   'rules': {
     'babel/no-await-in-loop': 0,
@@ -14,13 +13,5 @@ module.exports = {
     'flowtype/object-type-delimiter': 0,
     'no-new': 0,
     'linebreak-style': 0,
-    'prettier/prettier': [1, {
-      'bracketSpacing': false,
-      'jsxBracketSameLine': true,
-      'parser': 'flow',
-      'printWidth': 120,
-      'singleQuote': true,
-      'trailingComma': 'all',
-    }],
   },
 };
--- a/.flowconfig
+++ b/.flowconfig
@@ -10,4 +10,4 @@ flow-typed/
 experimental.const_params=true
 
 [version]
-0.57.3
+0.59.0
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ node_modules
 npm-debug.log
 build/
 flow-typed/
+package-lock.json
+yarn.lock
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,59 @@
+## v0.6.6
+
+- Allow filtering for didChangeWatchedFiles to workaround language servers triggering full rebuilds on changes to .git/build folders etc.
+- Add flow type definitions for Atom IDE UI's busy signal
+
+## v0.6.5
+
+- Send rootUri along with rootPath on initialize for compatibility with LSP v3 servers
+- New signature helper adapter for `textDocument/signatureHelp`
+- Upgrade various npm runtime and dev dependencies
+- Revert to using item.label when no item.insertText for AutoComplete+
+- Take priority over built-in AutoComplete+ provider
+
+## v0.6.4
+
+- Capture error messages from child process launch/exit for better logging
+- New `workspace/applyEdit` adapter
+- New `document/codeAction` adapter
+- Order OutlineView depending on source line & position
+
+## v0.6.3
+
+- Additional error logging
+
+## v0.6.2
+
+- Clear linter messages on shutdown
+
+## v0.6.1
+
+- Accidental republish of v0.6.0
+
+## v0.6.0
+
+- Handle duplicate change events for incremental doc syncing
+- Handle files opened multiple times in different windows/editors
+- Fix GitHub repo link in package.json
+- Ensure child process killed on exit/reload
+- Do not convert http:// and https:// uri's as if they were file://
+
+## v0.5.0
+
+- Allow duplicate named nodes in OutlineView
+- Do not npm publish pre-transpiled sources or misc files
+- Send LSP `exit` notification after `shutdown`
+- Use `atom.project.onDidChangeFiles` for fs monitoring instead of faking on save
+
+## v0.4.1
+
+- New `document/codeHighlights` adapter
+- Change nuclide flowtypes to atomIde
+- Remove redundant log messaging
+- Add eslint to build and make files compliant
+
+
+## v0.4.0
+
+- Switch code format to new range provider
+- Remove postInstall now project is released
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## v0.6.7
+
+- Update vscode-jsonrpc from 3.3.1 to 3.4.1
+- Allow file: uri without // or /// from the server
+
 ## v0.6.6
 
 - Allow filtering for didChangeWatchedFiles to workaround language servers triggering full rebuilds on changes to .git/build folders etc.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## v0.7.0
+
+- Support snippet type completion items
+- Move completionItem detail to right for consistency with VSCode
+- Make ServerManager restartable
+- Sort completion results
+- LSP v3 flow types plus wiring up of willSave
+- Support TextDocumentEdit in ApplyEditAdapter for v3
+- Upgrade flow, remove prettier
+- Busy Signals added for start and shutdown
+- Dispose connection on server stop, prevent rpc errors in console
+
 ## v0.6.7
 
 - Update vscode-jsonrpc from 3.3.1 to 3.4.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v0.7.2
+
+- AutoComplete to CompletionItems now actually work on Atom 1.24 not just a previous PR
+
 ## v0.7.1
 
 - AutoComplete to CompletionItems now support resolve when using Atom 1.24 or later
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v0.7.1
+
+- AutoComplete to CompletionItems now support resolve when using Atom 1.24 or later
+
 ## v0.7.0
 
 - Support snippet type completion items
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ The language server protocol consists of a number of capabilities. Some of these
 | workspace/didChangeWatchedFiles | Atom file watch API           |
 | textDocument/publishDiagnostics | Linter v2 push/indie          |
 | textDocument/completion         | AutoComplete+                 |
-| completionItem/resolve          | AutoComplete+      (TBD)      |
+| completionItem/resolve          | AutoComplete+ (Atom 1.24+)    |
 | textDocument/hover              | Atom-IDE data tips            |
 | textDocument/signatureHelp      | Atom-IDE signature help       |
 | textDocument/definition         | Atom-IDE definitions          |
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ The language server protocol consists of a number of capabilities. Some of these
 | window/showMessageRequest       | Notifications package         |
 | window/logMessage               | Ignored                       |
 | telemetry/event                 | Ignored                       |
-| workspace/didChangeWatchedFiles | [Atom file watch api](https://github.com/atom/atom/pull/14853) (TODO)    |
+| workspace/didChangeWatchedFiles | Atom file watch API           |
 | textDocument/publishDiagnostics | Linter v2 push/indie          |
 | textDocument/completion         | AutoComplete+                 |
 | completionItem/resolve          | AutoComplete+      (TBD)      |
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ Some more elaborate scenarios can be found in the [ide-java](https://github.com/
 
 Right now we have the following experimental Atom LSP packages in development. They are mostly usable but are missing some features that either the LSP server doesn't support or expose functionality that is as yet unmapped to Atom (TODO and TBD in the capabilities table above).
 
-### Atom team packages
+### Official packages
 
 * [ide-csharp](https://github.com/atom/ide-csharp) provides C# support via [Omnisharp (node-omnisharp)](https://github.com/OmniSharp/omnisharp-node-client)
 * [ide-flowtype](https://github.com/flowtype/ide-flowtype) provides Flow support via [Flow Language Server](https://github.com/flowtype/flow-language-server)
@@ -136,10 +136,7 @@ Right now we have the following experimental Atom LSP packages in development. T
 
 ### Community packages
 
-* [ide-reason](https://github.com/zaaack/atom-ide-reason) provides OCaml and Reason support using [OCaml Language Server](https://github.com/freebroccolo/ocaml-language-server)
-* [ide-rust](https://github.com/mehcode/atom-ide-rust) provides Rust support using [Rust Language Server](https://github.com/rust-lang-nursery/rls)
-* [ide-vue](https://github.com/rwatts3/atom-ide-vue) provides Vue language support using the [Vue Language Server](https://www.npmjs.com/package/vue-language-server)
-* [ide-python](https://github.com/lgeiger/ide-python) provides Python support via [Python Language Server](https://github.com/palantir/python-language-server)
+Our [full list of Atom IDE packages](https://github.com/atom/atom-languageclient/wiki/List-of-Atom-packages-using-Atom-LanguageClient) includes the community packages.
 
 ### Other language servers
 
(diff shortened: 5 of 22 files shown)


damieng:
Why did you close this? It looks super interesting!

deankevorkian:
Thanks! :) I closed this because I wanted to rethink a few things...
Watching the currentSuggestions via the `AutoLanguageClient` class is a bad idea since all open editors reach the same language client. I think this might be reproducable @ ide-typescript too. (I opened 2 editors, and the fuzzy filtering was done in the same list instead of keeping 2 seperate lists).
I plan to use a TextEditor to currentSuggestions Map object, and I'll reopen it when I have something more concrete.

Also for handling backspaces, I think we should handle this on any case of removing, not only backspace. So I plan on registering to the TextBuffer's `onDidChangeText` event and reset the `currentSuggestions` on each remove.

damieng:
We're also working on getting better support for trigger characters into autocomplete-plus itself. However we'll still need to cache results in our provider for things like resolve to work.

deankevorkian:
Reopening the PR - I now use a Map object to link each TextEditor to it's own `EditorSuggestionInfo` private type that has the atom autocomplete+ suggestions and an "isComplete" indication.

Scenario where I get the same suggestions (and same fuzzy-filter) for 2 different tabs/Editors is now fixed! :)
Now proceeding to work with TextBuffer to reset "isComplete" correctly for now..

leroix:
@deankevorkian I really like where this is going! I think it solves the performance issues with language servers really well!

I had an idea for simplifying the issue with backspace/character deletion.

1) If `request.prefix` is a trigger character, we fetch suggestions from the language server.
2) If the character preceding `request.prefix` is a trigger character and `isComplete` is `false`, we fetch suggestions from the language server.
3) We only return suggestions from `getSuggestions` when the character preceding `request.prefix` is a trigger character or if `request.prefix` is a trigger character.

I believe this would obviate the need for the `onDidChangeText` handler.

Here's how to check if the preceding character is a trigger character:
```
triggerChars.includes(request.editor.getBuffer().getTextInRange([
  {
    row: request.bufferPosition.row,
    column: request.bufferPosition.column - request.prefix.length - 1
  },
  {
    row: request.bufferPosition.row,
    column: request.bufferPosition.column - request.prefix.length
  }
]))
```

Let me know if I can clarify anything or help in any other way to get this merged. :)

deankevorkian:
@leroix Thanks!
It will simplify things (though checking if the character preceding request.prefix is a trigger character isn't enough, we can have incomplete suggestions by the language server after typing more than 1 character after the trigger char). 
Problem is it still doesn't provide a valid way of knowing when to reset isComplete and currentSuggestions. It's only effective when deleting a single character with a backspace.
Say we had `document.cha`, and the user removed `cha`, if he keeps typing we're going to think isComplete is true and we'll use fuzzy filtering on the suggestions (which are filtered to `document.cha*`). 
The only reason I use the onDidChangeText handler is to make sure we reset those indications on **any** type of deletion. :(

leroix:
@deankevorkian ah, I see my oversight there. Good point!

deankevorkian:
@damieng can we get this merged? :crossed_fingers:

damieng:
I'd really like it if we moved the functionality into the AutoCompleteAdapter rather than the AutoLanguageClient as the latter is getting quite large and cumbersome with all the things it does.  Thoughts?

leroix:
@deankevorkian does this still call `/completion` anytime a deletion occurs?

leroix:
Is it possible that the caching belongs in `AutocompleteAdapter` while the logic about when to actually return suggestions (i.e. trigger characters) still belongs in `AutoLanguageClient`. That way ide-* packages can override it easily?

deankevorkian:
@damieng I thought about this for a long time, when I started I thought of less edge cases and thought it would fit as part of the `AutoLanguageClient` itself. I thought it might be better to move the caching management (as @leroix suggested) to `AutocompleteAdapter` and leave the functionality of what/when/why/how to work with the cached suggestions in it's current place.

@leroix I was going to write 10 lines of why this calls /completion anytime a deletion occurs... but it doesn't :sweat: 
It does in call cases, except of a case of, say, having `document.char` and the user removes the last character ('r') - even though I do reset the cache and set `isComplete` to `false`, when I call getSuggestion after that deletion I'm gonna look for one of the `triggerCharacters`. Problem is I check that on the prefix. I'll need to handle that case by looking for a `triggerCharacter` **anywhere** in the buffer (and not just the prefix).
I'll work that case tomorrow and move the cache handling to the Adapter

deankevorkian:
This seems more difficult than I thought :disappointed: 
I need some help to correctly detect the case I described above. the request's `prefix` doesn't give much to handle such cases. Do we want to call `textDocument/completion` after every removal, on **any** case? (Removing a character, removing a line, etc)?
If so I can add an `isAfterRemoval` indication to the `EditorSuggestionInfo` (I'll rename it to `EditorSuggestionState`) - and check on that before all the other checks.

leroix:
@deankevorkian I don't think we need to call `textDocument/completion` after any sort of removal. I was just asking about the comment you had made earlier:

>Handle issue where deleting something always causes a /completion call (Add a check to see if we're past-trigger character before calling completion?)

Looking at your code, I don't think this is actually an issue. It doesn't seem to be calling /completion after every removal.

>I need some help to correctly detect the case I described above.

Can you elaborate a bit more on the case you're having trouble with? I'm not quite sure I follow.

deankevorkian:
@leroix Oh yes, it's not an issue. there's no /completion call after every removal at the moment.

The cases I'm having trouble with are cases after a removal was made. After a removal is done we clear the suggestion info (`currentSuggestions` to `[]` and `isComplete` to `false`). But the next time I'll type any character, I'm going to run the same logic - which, in case it wasn't a trigger character, will not reach the language server.

example: after typing `document.cha`, we're going to have a list of suggestions like this - 



But after removing the last `a` character, we're going to have a list of suggestions like this - 


Where we're actually expecting this...

damieng:
I think we should probably read backwards from the current cursor position
backward to the first non-alphanumeric character.  If it's a '.' (or rather
any of triggerCharacters from the initializationResponse) then we trigger
the autocomplete.

leroix:
@deankevorkian I think this is because you're replacing currentSuggestions with the fuzzy filtered results.

I think it will help to think about this as a caching problem where we're caching the results from /completion. We have to think hard on when to cache and what situations invalidate the cache. Here's an idea I was sketching out for the `getSuggestions` method of the language client before we started talking about moving the caching into the adapter. Note: I haven't gotten it to type-check yet (not super familiar with flow).

```javascript
async getSuggestions(request: atom$AutocompleteRequest): Promise<Array<atom$AutocompleteSuggestion>> {
    if (this.autoComplete == null) {
      this.autoComplete = new AutocompleteAdapter();
    }

    const server: ?ActiveServer = await this._serverManager.getServer(request.editor);
    if (server == null || !AutocompleteAdapter.canAdapt(server.capabilities)) {
      return Promise.resolve([]);
    }

    if (request.activatedManually) {
      return this.autoComplete.getSuggestions(server.connection, request).then(response => {
        return response.completionItems;
      });
    }

    if (request.prefix === '') {
      return Promise.resolve([]);
    }

    let triggerChars: Array<string>;

    if (server.capabilities.completionProvider && server.capabilities.completionProvider.triggerCharacters) {
      triggerChars = server.capabilities.completionProvider.triggerCharacters;
    } else {
      triggerChars = ['.'];
    }

    let prefixPosition: atom$PointObject;

    if (triggerChars.includes(request.prefix)) {
      prefixPosition = request.bufferPosition;
    } else {
      prefixPosition = {
        row: request.bufferPosition.row,
        column: request.bufferPosition.column - request.prefix.length
      };
    }

    let triggerChar: string = request.editor.getBuffer().getTextInRange([
      [prefixPosition.row, prefixPosition.column - 1],
      prefixPosition
    ]);

    if (!triggerChars.includes(triggerChar)) {
      return Promise.resolve([]);
    }

    let completionCache: ?CompletionResponseCache = this._completionCachesByEditor.get(request.editor);
    if (completionCache == null) {
      completionCache = {
        prefixPosition: prefixPosition,
        prefixes: new Map(),
      };
    }

    if (
      completionCache.prefixPosition.row === prefixPosition.row &&
      completionCache.prefixPosition.column === prefixPosition.column
    ) {
      let resultsForPrefix: ?AutocompletionResponse = completionCache.prefixes.get(request.prefix);

      let k: number = request.prefix.length;
      while (!resultsForPrefix && k >= 0) {
        k--
        let _resultsForPrefix: AutocompletionResponse = k === 0
          ? completionCache.prefixes.get(triggerChar)
          : completionCache.prefixes.get(request.prefix.slice(0,k));
        if (_resultsForPrefix.isComplete) {
          resultsForPrefix = _resultsForPrefix;
        }
      }

      if (resultsForPrefix) {
        return new Promise(resolve => {
          const filtered = filter(resultsForPrefix.completionItems, request.prefix, {key: 'text'}).map(s => 
          ({
            ...s,
            ...{replacementPrefix: request.prefix},
          }));
          resolve(filtered);
        })
      }
    }

    if (
      completionCache.prefixPosition.row !== prefixPosition.row ||
      completionCache.prefixPosition.column !== prefixPosition.column
    ) {
      completionCache = {
        prefixPosition: prefixPosition,
        prefixes: new Map(),
      };
      this._completionCachesByEditor.set(request.editor, completionCache);
    }

    return this.autoComplete.getSuggestions(server.connection, request).then(response => {
      completionCache.prefixes.set(request.prefix, response);
      return response.completionItems;
    });
  }
```

I also think https://github.com/facebook/nuclide/blob/master/pkg/commons-atom/AutocompleteCacher.js is a really good source of inspiration. I like how they encapsulated all of the caching logic.

Disclaimer: I haven't tested this, so take it with a grain of salt. I just thought maybe it could be a source of ideas.

deankevorkian:
@leroix this looks really good. One of the main things I was missing was finding a way getting to the last trigger character, which you've done in `column: request.bufferPosition.column - request.prefix.length`.
Thanks for the offer, I'll dive into it, as well as the AC Cacher the people @ Nuclide did. :bowing_man:

leroix:
@deankevorkian glad it helped! Thanks for continuing to push this PR forward :). 💯

damieng:
I think we're going to want to make sure most of this code lives in the AutoCompleteAdapter. AtomLanguageClient is getting very large and unfocused as we go on.

What's the issue with the tests?

Thanks for your help on this.

deankevorkian:
@damieng First thing I wanted to do was to rearrange the way we cache and return results, following @leroix's suggestions. Next thing I'll do is move all the caching logic to the `AutoCompleteAdapter` (while keeping the when-to-fetch/when-to-get-cached-suggestions logic on the `AutoLanguageClient` itself, to make it easier for packages to override.

@leroix  I finished implementing your overall idea, looks good and runs well, added some comments to clarify it to whoever sees (and for myself as well...). I'm still a bit unclear about the variable names. I think this creates a confusion between what is a prefix and what is a trigger character (as `prefixPosition` represents the last position of a prefix character from which we fetched the Language Server. We call it `prefixPosition` but maybe `lastTriggerCharPos` fits more, as we only fetch the LS for trigger characters, except from manual activation).
Still a little concerned about performance. There are also many redundant nulls I have to re-check in the code because of Flow. Hope it's still ok this way.

leroix:
@deankevorkian this looks great. I'm hopefully going to get a chance to review it in detail in the next day or two.

deankevorkian:
I'm still concerned we're not covering all cases where the cache should be reset.
We now do that by simply checking on the prefix position and check if it changed.
We're gonna have to track: 
1. if the `triggerCharacter` changed on the last saved position (the one at `prefixPosition`) then reset the cache. (because we may have multiple trigger characters for a LS but we only check on the position before we return cached results)

2. reset the cache if text was changed anywhere **before** the `triggerCharacter`. Example -
replace `Array.` with `Error.` - the triggercharacter (`.`) remains on the same `prefixPosition` - so we'll unwillingly return the results for `Array.` instead of `Error.`.

I'll be able to handle case 1 by adding a `characterAtPos` or `triggerCharacter` to `CompletionCache` (then, I'll check if the prefix position has changed OR the trigger character has changed).
And case 2......... I'll probably have to handle by (again... 😞) registering to the editor's Buffer and see if a change has occured **before** the prefixPosition. I know we probably don't want this, but I'll do it anyway and tell me what you think..

This may be a bit confusing - I'll try to provide pictures of the 2 mentioned cases when I have some time

deankevorkian:
@leroix, out of what I wrote above, I (hope) I handled case 1 by including the trigger character with the prefixes (while reading/storing with the cache, I save the trigger-char & prefix, concatenated). This way I return cached values according to the trigger character (document.bla and document:bla return different results now). So the `prefixes` `map` keys look like this:
`.bla`, `:bla`
instead of just `bla` without the trigger character. this let's me return cached results according to the trigger char itself.

I still need help with number 2 - detecting if the text was changed **before** the cached triggerPosition. I could do it via the `TextBuffer`'s `onDidChangeText` event, but maybe you have a better idea.
After we get done with 2, I think we're good.

leroix:
@deankevorkian Sorry for the delay on this. I'm checking out your branch and trying it out. It looks good so far. I'm not opposed to using `onDidChangeText` if it turns out that we need it. I just want to make sure we're implementing the simplest possible solution.

I'll report back as I start to understand what you've done so far a little better.

damieng:
Anything outstanding on this other than the merge conflicts?  I'd love to get it put in so I can rebase my completionItem/resolve work over the top without further causing you conflicts.

leroix:
@deankevorkian I think case 2 is kind of an edge case, but perhaps we could support it by letting the cache key be `<word><trigger character><prefix>` e.g. `document.bla`. I'm not sure if it's worth it though.

Other than that, I think this PR looks pretty good. It's likely that things will continue to evolve, so it seems better to get something out there sooner rather than later.

deankevorkian:
I agree, we should probably merge the changes. Other issues will probably come up later as people start using it, so it's better to get something out there. Other than case 2 I'm ok with the changes, I think it's ready :)

deankevorkian:
Done.

damieng:
Sorry, I just conflicted your PR again! This one is going to be tricky - don't worry I'll take care of merge integration. I think we also want to reduce the cache count for sake of complexity.

deankevorkian:
Since we clear the cache everytime the prefix character's position changes or the position remains the same - but the prefix character changes. I'm not sure, given the current caching logic that this is something we should worry about. Usually, most of the time, the cache will contain 1 item. it will contain more than one item only if one of the conditions above doesn't occur.
@leroix since you've been working on performance improvements @ autocomplete lately, and since I followed your idea, what's your opinion on this?

leroix:
@deankevorkian I think @damieng's concerns are more to do with the cyclomatic complexity of the code rather than the amount of data that gets cached. I think he's right in that we should find the simplest solution that performs well.

damieng:
The main thing is I don't think we need to cache for multiple texteditors. The user is only interacting with one and just make a note of which one we cached for.  Also with the new new resolve code we have to have a map of the items as they came back from the LSP to their autocomplete equivalents so we can get the original LSP item required for the LSP resolve api call.

damieng:
I apologize for bringing this up late in the PR but I'm more than happy to make the changes.

deankevorkian:
Oh I see, gotcha. I'll get to it then. I'll start with merging master to my branch to get the new resolve code, and change the caching logic & structure to only hold cache for the current editor.

damieng:
Thanks! I'll merge this in if you're ready.

deankevorkian:
Done!

deankevorkian:
@damieng was this PR reset/reverted? I started working on a PR that takes into account using some of the logic & types. I want to know if that's the case in order to decide if I should start working on it from scratch or continue.

damieng:
Hey @deankevorkian 

I had to revert it as I discovered a bunch of issues after I merged and even after several hours couldn't unwrap them all.  Some of them were caused by the integration with the resolve stuff, others were not (like manual trigger using the wrong cache) and I needed to get some changes out for getting the autocomplete in ide-typescript out that were blocked by this.

I wanted to let you know sooner but was trying to figure out a way to go forward.  Right now I think going with a slight modification of the code in ide-typescript might be the best option and to keep the caching super-simple because of all these edge cases where it can get out of sync.

deankevorkian:
I realized that's the case. Sorry for the trouble! I'll try getting your
changes into my branch and see if I can make it work with the resolve
stuff. Could you let me know what other cases you found so I can work them
too? Manual triggering isn't supposed to use the cache, it should call the
LS instead. I'll check that too.

On Dec 13, 2017 22:23, "Damien Guard" <notifications@github.com> wrote:

Hey @deankevorkian <https://github.com/deankevorkian>

I had to revert it as I discovered a bunch of issues after I merged and
even after several hours couldn't unwrap them all. Some of them were caused
by the integration with the resolve stuff, others were not (like manual
trigger using the wrong cache) and I needed to get some changes out for
getting the autocomplete in ide-typescript out that were blocked by this.

I wanted to let you know sooner but was trying to figure out a way to go
forward. Right now I think going with a slight modification of the code in
ide-typescript might be the best option and to keep the caching
super-simple because of all these edge cases where it can get out of sync.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/atom/atom-languageclient/pull/128#issuecomment-351511080>,
or mute the thread

deankevorkian:
@damieng should I start a new PR for sake of cleanness or continue here? Also could you please tell me what other cases weren't working properly?

damieng:
I want to get the super-simple reduced caching version of the autocomplete triggers in to this from ide-typescript.  Once that's done if you want to look into applying the cache magic on top that would be great.  I can ping you once it's in.

PR #129, opened 2017-10-30
Fixed file URI parsing

laughedelic:
Double slash after the `file:` protocol name is optional and shouldn't be required. See for example this [SO answer](https://stackoverflow.com/a/44725349/736957) with references to the [URI specification](https://tools.ietf.org/html/rfc3986#section-3).

I discovered this problem when the server (for which I'm writing a plugin) sent `publishDiagnostic` with URI like `file:/root/path/file.foo`. Atom treated this as a relative path, so it became `/root/path/file:/root/path/file.foo`.

I'm totally new to Electron, JavaScript, etc. So this may be not the best solution. Tell me if I need to change it.

### DIFF
--- a/lib/convert.js
+++ b/lib/convert.js
@@ -24,15 +24,15 @@ export default class Convert {
   // * `uri` A Uri to convert to a file path.
   //
   // Returns a file path corresponding to the Uri. e.g. /a/b/c.txt
-  // If the Uri does not begin file:// then it is returned as-is to allow Atom
+  // If the Uri does not begin file: then it is returned as-is to allow Atom
   // to deal with http/https sources in the future.
   static uriToPath(uri: string): string {
-    let filePath = decodeURIComponent(uri);
-    if (!filePath.startsWith('file://')) {
+    const url = require('url').parse(uri);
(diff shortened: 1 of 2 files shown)


damieng:
Traditionally the file url has required /// so while I'm okay with taking this to be more flexible with the input we receive you'll find plenty of other consumers won't handle it properly so you might want to go with file:/// for better backwards compatibility.

damieng:
Thanks!

laughedelic:
> you'll find plenty of other consumers won't handle it properly so you might want to go with file:/// for better backwards compatibility

I thought about it and I already notified server developers. The URI formatted without slashes [in that server](https://github.com/sbt/sbt/blob/v1.1.0-M1/main/src/main/scala/sbt/internal/server/LanguageServerReporter.scala#L91) comes from the standard Java `file.toURI`, so I would assume that it's quite common. Also it seems that VS Code recognizes such URIs without any problem. Anyway, I think the main point of this change was to use URL parsing instead of matching on prefix and substring cutting.

laughedelic:
@damieng Could you make a new release with this change?

Issue #31, opened 2017-05-02
Implement and Advertise Snippet Support


wbinnssmith:
Issue-izing the TODO found at [the autocomplete adapter](https://github.com/atom/atom-languageclient/blob/2b3ac65e87cb2832726e61502f1a2ab9d601fe6f/lib/adapters/autocomplete-adapter.js#L32)

When snippet support is implemented, the client should should also advertise this via the client capabilities sent to each language server.

damieng:
Right now we're targeting v2 of the spec (v3 is still in flux) so we can skip the client capabilities part for now.

wbinnssmith:
oh! didn't even realize client capabilities was part of v3.

it looks like v2 has a distinct `Snippet` completion type (which v3 has as well), but v3 (in its current form at least) has the option to use a snippet as the `insertTextFormat` for non-Snippet completion types. Looks like that's where having the client advertise its support for snippets is most useful as the server can decide the appropriate `insertTextFormat` to send back.

In the case of Flow, we can provide function completions with snippet text that expand to fill in the function's arguments, so this will be useful when we support v3.

rcjsuen:
> v3 is still in flux

I thought v3 came out in [December 2016](https://github.com/Microsoft/language-server-protocol/commit/164eb31bd49535dca046034602146f96bb17b56d). Am I misunderstanding something?

damieng:
https://github.com/Microsoft/language-server-protocol/commits/master/protocol.md shows there were changes all through May.

We'll add support for the v3 features in a future update.

rcjsuen:
I don't believe those are breaking changes but certainly better safe than sorry.

damieng:
v2 part is now done.

damieng:
@wbinnssmith I think this has now been implemented through various incremental commits - can you confirm?

UziTech:
Development of [atom-languageclient](https://www.npmjs.com/package/atom-languageclient) has officially moved to https://github.com/atom-ide-community/atom-languageclient 🎉 

If this is still an issue please consider opening an issue on that repo.

PR #132, opened 2017-11-02
Support snippets in autocomplete suggestions

hansec:
Add support for snippets in completion items returned by language servers.

### DIFF
--- a/lib/adapters/autocomplete-adapter.js
+++ b/lib/adapters/autocomplete-adapter.js
@@ -92,7 +92,7 @@ export default class AutocompleteAdapter {
   //
   // Returns an AutoComplete+ suggestion.
   static basicCompletionItemToSuggestion(item: CompletionItem): atom$AutocompleteSuggestion {
-    return {
+    var sugg = {
       text: item.insertText || item.label,
       displayText: item.label,
       filterText: item.filterText || item.label,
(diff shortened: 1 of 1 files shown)


lib/adapters/autocomplete-adapter.js
```diff
@@ -101,6 +101,10 @@ export default class AutocompleteAdapter {
       description: item.documentation,
       descriptionMarkdown: item.documentation,
     };
+    if (item.insertTextFormat == 2) {
+      sugg.snippet = item.insertText;
+    };
+    return sugg;
```
damieng: Thanks for this! Could we maybe go with just a 1-line change version of having the snippet added, e.g.

snippet: item.insertTextFormat == 2 ? item.insertText : undefined

damieng:
Fixes #31

hansec:
Simplified as requested. Sorry for the verbosity, Javascript is not my primary language.

lib/adapters/autocomplete-adapter.js
```diff
@@ -96,6 +96,7 @@ export default class AutocompleteAdapter {
       text: item.insertText || item.label,
       displayText: item.label,
       filterText: item.filterText || item.label,
+      snippet: item.insertTextFormat == 2 ? item.insertText : undefined,
```
damieng: Sorry, my mistake. It's supposed to be `===` to keep the linter happy.  (JS not the language I'm most familiar with either ;-)  )

PR #144, opened 2017-11-16
Add Language Server Protocol v3 support

daviwil:
This change updates our LSP types and client interface to conform to LSP v3.  I've also added support for sending the `textDocument/willSave` event and applying `TextDocumentEdits` if those are sent in an `workspace/applyEdit` request.

I haven't yet added version checking to the code that applies TextDocumentEdits, planning to do that a bit later when I'm back home.  Will add a test or two for that also.

**Questions for reviewers**

1. For the LSP message types, should we keep the same ordering as what they have in the protocol.md file just to make it easier to spot-check protocol changes in the future?
2. Looks like there's some comment block inconsistency in the LSP types, probably due to copy/pasting types from the LSP spec over time.  Should I clean those up to be consistent?
3. Any other tests you think I should add?

### DIFF
--- a/lib/adapters/document-sync-adapter.js
+++ b/lib/adapters/document-sync-adapter.js
@@ -3,6 +3,7 @@
 import {
   LanguageClientConnection,
   FileChangeType,
+  TextDocumentSaveReason,
   TextDocumentSyncKind,
   TextDocumentSyncOptions,
   type TextDocumentContentChangeEvent,
@@ -160,6 +161,7 @@ class TextEditorSyncAdapter {
     }
 
     this._disposable.add(
+      editor.getBuffer().onWillSave(this.willSave.bind(this)),
       editor.onDidSave(this.didSave.bind(this)),
       editor.onDidDestroy(this.didClose.bind(this)),
       editor.onDidChangePath(this.didRename.bind(this)),
@@ -296,6 +298,18 @@ class TextEditorSyncAdapter {
     }
 
     this._connection.didCloseTextDocument({textDocument: {uri: this.getEditorUri()}});
+  }
+
+  // Called just before the {TextEditor} saves and sends the 'willSaveTextDocument' notification to
+  // the connected language server.
+  willSave(): void {
+    if (!this._isPrimaryAdapter()) return;
+
+    const uri = this.getEditorUri();
+    this._connection.willSaveTextDocument({
+      textDocument: {uri},
+      reason: TextDocumentSaveReason.Manual
+    });
   }
 
   // Called when the {TextEditor} saves and sends the 'didSaveTextDocument' notification to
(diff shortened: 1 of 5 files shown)


lib/auto-languageclient.js
```diff
@@ -81,7 +81,22 @@ export default class AutoLanguageClient {
...
+          workspaceEdit: {
+            documentChanges: true
+          }
+        },
+        textDocument: {
+          synchronization: {
+            dynamicRegistration: false,
+            willSave: true,
+            willSaveWaitUntil: false,
+            didSave: true
```
daviwil: Instead of sending over only what wouldn't be default values, should I be explicit and send the whole block even if most of them will have false values?

lib/languageclient.js
```diff
@@ -374,7 +410,246 @@ export type InitializeParams = {
...
+	 * The client supports applying batch edits to the workspace by supporting
+	 * the request 'workspace/applyEdit'
+	 */
+	applyEdit?: boolean;
+
+	/**
+	 * Capabilities specific to `WorkspaceEdit`s
+	 */
+	workspaceEdit?: {
+		/**
```
daviwil: Looks like I need to clean up indentation width in a lot of places.

lib/adapters/document-sync-adapter.js
```diff
@@ -160,6 +161,8 @@ class TextEditorSyncAdapter {
     }
 
     this._disposable.add(
+      editor.getBuffer().onWillSave(this.willSave.bind(this)),
+      editor.onDidSave(this.didSave.bind(this)),
```
damieng: Duplicate line?

daviwil: Ouch!

lib/auto-languageclient.js
damieng: I don't have a strong preference here - initialize is only called once so no perf cost to send them all.  I doubt they'd change default either.

lib/languageclient.js
```diff
@@ -129,6 +134,10 @@ export class LanguageClientConnection {
     this._sendNotification('textDocument/didClose', params);
   }
 
+  willSaveTextDocument(params: WillSaveTextDocumentParams): void {
```
damieng: Should probably add a Public doc comment for consistency with the other methods.

daviwil: Will do, I'll add comments on others that are missing them too.

lib/auto-languageclient.js
```diff
@@ -81,7 +81,22 @@ export default class AutoLanguageClient {
   getInitializeParams(projectPath: string, process: child_process$ChildProcess): ls.InitializeParams {
     return {
       processId: process.pid,
-      capabilities: {},
+      capabilities: {
+        workspace: {
+          applyEdit: true,
```
rictic: Consider also:

`didChangeWatchedFiles: true,`

lib/auto-languageclient.js
rictic: Small +1 to sending them all, as this would also be a good quick bit of documentation for what's implemented so far, and may also serve as a reminder to send up a `true` once a feature is implemented.

daviwil: Yep, planning to add the rest of them before merging!

PR #145, opened 2017-11-17
Changed to using rimraf instead of rm -rf for cross-platform support

deankevorkian:
As title suggests. `rm -rf` is unix-based. this pr switches it's change to `rimraf` package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "license": "MIT",
   "main": "./build/lib/main",
   "scripts": {
-    "clean": "rm -rf build",
+    "clean": "rimraf build",
     "compile": "babel lib --out-dir build/lib && babel test --out-dir build/test",
     "flow": "flow",
     "lint": "eslint lib test --max-warnings 0",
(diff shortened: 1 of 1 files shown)


damieng:
I'm not keen on adding a runtime dependency on rimraf to all ide-* packages just for the sake of being able to clean while developing on atom-languageclient.

deankevorkian:
Correct me if I'm wrong but I don't think it's a runtime dependency if we put it on devDependencies - it won't impact performance/load times for Atom users, only for developers?

damieng:
Unfortunately apm installs all devdependencies because it might need to compile/transpile etc. I'm pretty sure (but not 100%) that this also applies to downstream npm dependencies.

damieng:
rimraf unfortunately also brings in glob, which brings in  fs.realpath, inflight, inherits, minimatch, once, path-is-absolute which brings in...

damieng:
FWIW I'm Windows based but http://cmder.net/ gives me `rm` and other unixy tools.

deankevorkian:
Gotcha. I work from both, so I don't really mind - just stumbled across this while working on Windows, thought it'd be harmless to add this change. Thanks for the quick reply - closing this!

PR #148, opened 2017-11-22
Resolve implementation for new AutoComplete hook

### LINKED ISSUE #7: completionItem/resolve
Auto completion items may not be fully completed during the initial completion request and can be further resolved to fill in the missing details by calling [completionItem/resolve](https://github.com/Microsoft/language-server-protocol/blob/master/versions/protocol-2-x.md#completionItem_resolve).

## Challenges
- Autocomplete+ does not have a hook for this right now
- Autocomplete+ would need a new API for this
- Find a language server that requires this

Activate if `capabilities.completionOptions.resolveProvider === true`

damieng:
Fixes #7

### DIFF
--- a/flow-libs/atom.js.flow
+++ b/flow-libs/atom.js.flow
@@ -1763,6 +1763,9 @@ type atom$AutocompleteProvider = {
   getSuggestions: (
     request: atom$AutocompleteRequest,
   ) => Promise<?Array<atom$AutocompleteSuggestion>>,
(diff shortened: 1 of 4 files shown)


hansonw:
@ljw1004 it's happening already! :) we should be able to use this hook in Nuclide too.

paradoxxxzero:
It doesn't seem to respect the `capabilities.completionProvider.resolveProvider` option. It makes ide-python crash with a Method Not Found.

PR #149, opened 2017-11-24
Add busy-signals for server start, initialize & stop

### LINKED ISSUE #105: Use atom-ide-busy to indicate start-up and shut-down activity
As per the discussion in https://github.com/facebook-atom/atom-ide-ui/issues/18

alexheretic:
* Add support for consuming `BusySignalService`
* Add busy signal `Starting ${this.getServerName()} for ${path.basename(projectPath)}`
* Add busy signal `${this.getServerName()} initializing for ${path.basename(projectPath)}`
* Add busy signal `Stopping ${this.getServerName()} for ${path.basename(projectPath)}`

Note: Having public _busySignalService_ in `AutoLanguageClient` is intended as this service can be useful to the subclass clients _(and they have to add the package.json declaration anyway)_. For example, we'll use the service to show we're installing Rls.

atom-languageclient-busy-signal

Resolves #105

### DIFF
--- a/lib/auto-languageclient.js
+++ b/lib/auto-languageclient.js
@@ -3,6 +3,7 @@
 import * as cp from 'child_process';
 import * as ls from './languageclient';
 import * as rpc from 'vscode-jsonrpc';
+import path from 'path';
 import {CompositeDisposable, Disposable} from 'atom';
 import {ConsoleLogger, NullLogger, type Logger} from './logger';
 import {ServerManager, type ActiveServer} from './server-manager.js';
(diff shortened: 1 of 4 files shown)


lib/auto-languageclient.js
```diff
@@ -34,6 +35,16 @@ export default class AutoLanguageClient {
...
   _signatureHelpRegistry: ?atomIde$SignatureHelpRegistry;
 
+  /**
+   * Available if consumeBusySignal is called, ie the client package.json includes:
+   * ```json
+   * "consumedServices": {
+   *   "atom-ide-busy-signal": { "versions": { "0.1.0": "consumeBusySignal" } }
+   * }
+   * ```
+   */
```
damieng: I think we should remove this from here - we don't explicitly call out any of the other services that need to be added.  We'll instead add busy signal to the example starter project.

lib/server-manager.js
```diff
@@ -35,19 +35,25 @@ export class ServerManager {
   _startForEditor: (editor: atom$TextEditor) => boolean;
   _startServer: (projectPath: string) => Promise<ActiveServer>;
   _changeWatchedFileFilter: (filePath: string) => boolean;
+  _getBusySignalService: () => ?atomIde$BusySignalService;
+  _languageServerName: string;
   _isStarted = false;
 
   constructor(
+    languageServerName: string,
```
damieng: Perhaps move this parameter to the end of the list in case anyone is using this class directly.

damieng:
Great job, thanks for this!

lib/auto-languageclient.js
alexheretic: Ok I'll tone it down to just a note to help those that find it to be null

PR #154, opened 2017-11-27
Reorganize LSP message types, add missing registration types

daviwil:
This change reorganizes the language server protocol Flow type definitions
so that they follow the order of the LSP spec for easier spot-checking.
It also adds some missing registration options types which currently
aren't being used.


lib/languageclient.js
```diff
@@ -852,6 +880,186 @@ export type TextDocumentRegistrationOptions = {
   documentSelector: DocumentSelector | null,
 };
 
+/**
```
damieng: Nitpick: Comment format.

daviwil: Will fix!

PR #158, opened 2017-12-07
Add `handleServerStderr` method + default logging

### LINKED ISSUE #157: Logging server stderr
I noticed you've disabled this. This would be very useful for Rls debugging, and I imagine any language server. I've also seen better results logging stderr lines, rather than chunks. Would you accept a PR to enable this functionality? I think it should at least be opt-in by clients.

alexheretic:
As discussed a little in #157 this PR adds stderr handling to a new public method:
```js
/**
 * Called on language server stderr output.
 * @param stderr a chunk of stderr from a language server instance
 */
handleServerStderr(stderr: string, projectPath: string) {
  stderr.split('\n').filter(l => l).forEach(line => this.logger.warn(`stderr ${line}`));
}
```

* Restored stderr / childProcess error handling code.
* Sends stderr chunks to new method, default impl warns per line

Resolves #157

### DIFF
--- a/lib/auto-languageclient.js
+++ b/lib/auto-languageclient.js
@@ -232,7 +232,7 @@ export default class AutoLanguageClient {
     } finally {
       startingSignal && startingSignal.dispose();
     }
-
+    this.captureServerErrors(process, projectPath);
     const connection = new ls.LanguageClientConnection(this.createRpcConnection(process), this.logger);
     this.preInitialization(connection);
     const initializeParams = this.getInitializeParams(projectPath, process);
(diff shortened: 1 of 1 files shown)


lib/auto-languageclient.js
```diff
@@ -270,16 +269,6 @@ export default class AutoLanguageClient {
     });
   }
 
-  handleSpawnFailure(err: any): void {
```
damieng: I think we still need this to report when starting a language server fails to the user.

lib/auto-languageclient.js
```diff
@@ -572,4 +561,12 @@ export default class AutoLanguageClient {
...
     return true;
   }
+
+  /**
+   * Called on language server stderr output.
+   * @param stderr a chunk of stderr from a language server instance
+   */
+  handleServerStderr(stderr: string, projectPath: string) {
+    this.logger.debug(`stderr ${stderr}`);
+  }
```
damieng: I'm okay with us logging by default.  I wonder if we should chose .error ?

alexheretic: It was _warn_ in the unused code previously. I don't mind!

lib/auto-languageclient.js
alexheretic: How does this work though, node processes don't seem to have an `error` event as the previous (unused) code used to call this method.

lib/auto-languageclient.js
alexheretic: This should be for each line too, the console handles it much better.

alexheretic:
warn instead of error makes sense as stderr could be used for all sorts really.

lib/auto-languageclient.js
damieng: https://nodejs.org/api/child_process.html#child_process_event_error

alexheretic: Ah right thanks, I was looking at `process` rather than child-

alexheretic:
@damieng I've squashed the change down and I think we've addressed all the issues. Are there any further problems do you think?

Issue #140, opened 2017-11-08
Implement request cancellation ($/cancelRequest)


daviwil:
Spec: [`$/cancelRequest`](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#cancelRequest)

vscode-jsonrpc provides a [`CancellationToken`](https://github.com/Microsoft/vscode-languageserver-node/blob/master/jsonrpc/src/cancellation.ts#L16) type which we could potentially re-expose through atom-languageclient.  We'd have to wire up a CancellationToken parameter on `LanguageClientConnection._sendRequest` and add one as an optional parameter to any method that calls it.  Luckily the RPC library [takes care of](https://github.com/Microsoft/vscode-languageserver-node/blob/6a3986aecac325361db4cc8a137415fd3b8337cb/jsonrpc/src/main.ts#L969) sending the `$/cancelRequest` notification if a CancellationToken gets cancelled.

damieng:
I think we'd want to automatically cancel them in autolanguageclient when we get a subsequent request for another one that would make the previous one useless. e.g. autocomplete, outline view, hover, signature help, definitions.

Probably not for some others though like find all references.

daviwil:
Sounds good!  I suppose the `LanguageClientConnection` wrapper methods are the right place to do that?

damieng:
They'll need to return the cancellationToken but we'll probably need to keep the most recent for the ones that auto cancel either on the instance of the adapter or somewhere else that we can track on a per-server basis.

This might be time to make all the adapters server-instance instead of the current sharing etc. they do.

The alternative is to track them on AutoLanguageClient but I fear it's a bit of a dumping ground.

hansonw:
We've actually been thinking about adding some sort of cancellation token mechanism to atom-ide-ui's APIs so that they can just be directly forwarded over. Google just decided to implement abortable fetches in Chrome so it may not look exactly like VSCode's `CancellationToken`, but it should be easily adaptable. https://developers.google.com/web/updates/2017/09/abortable-fetch

cc @matthewwithanm

daviwil:
Status update: Damien and I decided to put this one on hold until we can decide on the right way to expose cancellation from the atom-languageclient APIs.  The choices are between exposing vscode-jsonrpc's CancellationToken type all the way up through the API and services (not that desirable) or implementing some form of CancellablePromise which wraps their CancellationToken.

PR #160, opened 2017-12-21
Add CancellationToken support to LanguageClientConnection

daviwil:
This change adds and extra parameter to LanguageClientConnection._sendRequest which accepts the vscode-jsonrpc library's CancellationToken class which can be used to cancel active requests.  When a request is cancelled using one of these tokens, a `$/cancelRequest` notification will be sent to the language server to request a cancellation, but the server must implement cancellation behavior for this to be effective.

This change also adds auto-cancellation of previous requests in the AutocompleteAdapter and OutlineViewAdapter classes.

Related to #140.

### DIFF
--- a/lib/adapters/outline-view-adapter.js
+++ b/lib/adapters/outline-view-adapter.js
@@ -2,11 +2,16 @@
 
 import {LanguageClientConnection, SymbolKind, type ServerCapabilities, type SymbolInformation} from '../languageclient';
 import Convert from '../convert';
+import Utils from '../utils';
 import {Point} from 'atom';
+import {CancellationTokenSource} from 'vscode-jsonrpc';
 
 // Public: Adapts the documentSymbolProvider of the language server to the Outline View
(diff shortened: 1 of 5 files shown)


daviwil:
Failed to clear the linting errors before submitting, will get that fixed.

lib/adapters/outline-view-adapter.js
```diff
@@ -27,9 +31,25 @@ export default class OutlineViewAdapter {
...
-    });
+    }, cancellationToken.token);
+
+    this._cancellationTokens.delete(connection);
+    if (cancellationToken.token.isCancellationRequested) {
+      return {
+        outlineTrees: [],
+      };
+    }
+
```
damieng: Wonder if we can combine this logic with the same logic in the autocomplete adapter into a reusable function/chunk?

daviwil: I'll give it a shot!

PR #162, opened 2017-12-29
Check server capabilities before sending completionItem/resolve request

### LINKED ISSUE #161: completionProvider.resolveProvider option is not respected
For visibility, the PR #148 makes ide-python crash with a Method Not Found in Atom 1.24-beta1 because [python-language-server](https://github.com/palantir/python-language-server/issues/213) does not support `completionItem/resolve` as it is stated in its [capabilities](https://github.com/palantir/python-language-server/blob/develop/pyls/python_ls.py#L37).

lgeiger:
This PR checks if a server supports `completionItem/resolve` requests before sending one.

@paradoxxxzero Can you confirm that this fixes #161?


lgeiger:
I added a commit to fix CI I'm happy to split this into another PR if you prefer.

damieng:
Thanks for this!

paradoxxxzero:
@lgeiger Yes it works :+1:

PR #167, opened 2018-01-08
Add configuration watching to send `workspace/didChangeConfiguration`

daviwil:
This change adds logic to AutoLanguageClient so that a language package can opt-in to having a specific configuration section watched for changes so that workspace/didChangeConfiguration notifications are sent to the language server.  Language packages can opt-in by implementing the `getRootConfigurationKey` method and optionally the `filterConfigurationObject` method to manipulate the configuration object before it gets sent to the language server.

### DIFF
--- a/lib/auto-languageclient.js
+++ b/lib/auto-languageclient.js
@@ -178,6 +178,16 @@ export default class AutoLanguageClient {
     return this.socket != null ? 'socket' : 'stdio';
   }
 
+  // Return the name of your root configuration key
+  getRootConfigurationKey(): string {
+    return '';
+  }
+
+  // Optionally transform the configuration object before it is sent to the server
+  mapConfigurationObject(configuration: any): any {
+    return configuration;
+  }
+
   // Default implementation of the rest of the AutoLanguageClient
   // ---------------------------------------------------------------------------
 
@@ -251,6 +261,20 @@ export default class AutoLanguageClient {
     };
     this.postInitialization(newServer);
     connection.initialized();
+
+    const configurationKey = this.getRootConfigurationKey();
+    if (configurationKey) {
+      this._disposable.add(
+        atom.config.observe(configurationKey, config => {
+          const mappedConfig = this.mapConfigurationObject(config || {});
+          if (mappedConfig) {
+            connection.didChangeConfiguration({
+              settings: mappedConfig,
+            });
+          }
+        }));
+    }
+
     this.startExclusiveAdapters(newServer);
     return newServer;
   }


lib/auto-languageclient.js
```diff
@@ -178,6 +178,16 @@ export default class AutoLanguageClient {
     return this.socket != null ? 'socket' : 'stdio';
   }
 
+  // Return the name of your root configuration key
+  getRootConfigurationKey(): string {
```
daviwil: Also considered the name `getConfigurationPath`, any better ideas?

lib/auto-languageclient.js
```diff
@@ -251,23 +261,36 @@ export default class AutoLanguageClient {
...
-    childProcess.stderr.on('data', (chunk: Buffer) => {
-      const errorString = chunk.toString();
-      this.handleServerStderr(errorString, projectPath);
-      // Keep the last 5 lines for packages to use in messages
-      this.processStdErr = (this.processStdErr + errorString)
-        .split('\n')
-        .slice(-5)
-        .join('\n');
-    });
+    if (childProcess.stderr) {
```
daviwil: This was needed because of the way I'm launching my language server in ide-powershell.  I don't actually use ChildProcess for this, I use node-pty instead.  As far as I know I can't get an actual ChildProcess object from there so this bit of code needed a guard to avoid trouble with stdio streams not being available.

daviwil:
Argh, forgot to lint again :/

lib/auto-languageclient.js
daviwil: Decided to drop this change and create my own "fake" ChildProcess which conforms to what atom-languageclient expects.

lib/auto-languageclient.js
```diff
@@ -178,6 +178,16 @@ export default class AutoLanguageClient {
     return this.socket != null ? 'socket' : 'stdio';
   }
 
+  // Return the name of your root configuration key
+  getRootConfigurationKey(): string {
+    return '';
+  }
+
+  // Optionally filter/transform the configuration object before it is sent to the server
+  filterConfigurationObject(configuration: any): any {
```
damieng: Should probably call this map or transform as it will likely be used to reshape?

lib/auto-languageclient.js
```diff
@@ -251,6 +261,17 @@ export default class AutoLanguageClient {
     };
     this.postInitialization(newServer);
     connection.initialized();
+
+    const configurationKey = this.getRootConfigurationKey();
+    if (configurationKey) {
+      this._disposable.add(
+        atom.config.observe(configurationKey, config => {
+          connection.didChangeConfiguration({
+            settings: this.filterConfigurationObject(config || {}),
```
damieng: Might be worth moving the filter/map outside the connection.onDidChangeConfiguration call and test it for null first so that people can entirely filter out sending stuff if it turns out there's nothing they want to send.

daviwil: Good point!

lib/auto-languageclient.js
daviwil: Yep, filter is the wrong word, i'll change it to "map"

daviwil:
OK to merge this?

PR #169, opened 2018-01-13
Gracefully handle unexpected server shutdown

daviwil:
This change adds some logic to gracefully handle sudden server shutdown
like when the language server process gets killed by the user in their
task manager.  The solution is to watch for the RPC connection to get
closed unexpectedly and then dispose of the server in ServerManager
so that no further requests are sent to it.

### DIFF
--- a/lib/languageclient.js
+++ b/lib/languageclient.js
@@ -1,19 +1,29 @@
 // @flow
 
 import * as jsonrpc from 'vscode-jsonrpc';
+import EventEmitter from 'events';
 import {NullLogger, type Logger} from './logger';
 
 // Flow-typed wrapper around JSONRPC to implement Microsoft Language Server Protocol v3
 // https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md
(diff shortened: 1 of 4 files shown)


daviwil:
Happy to modify this solution if you think there's a better way to approach it!  I know that the server reloading refactoring will probably find another way to do all this, but this fix is helpful to me in the short term.

lib/languageclient.js
```diff
@@ -1,19 +1,28 @@
...
   constructor(rpc: jsonrpc.connection, logger: ?Logger) {
+    super();
     this._rpc = rpc;
     this._log = logger || new NullLogger();
     this.setupLogging();
     rpc.listen();
+
+    this.isConnected = true;
+    this._rpc.onClose(() => {
+      this.isConnected = false;
```
daviwil: Maybe I should add a log statement here?  What level?

daviwil:
Any feedback on this?  If not I'll just merge it this afternoon.

PR #171, opened 2018-01-18
Send non-null `initialized` params

alexheretic:
This pr sends non-null params with `initialized` notifications.

According to the [spec](https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md#initialized-notification-arrow_right) the _initialized_ parameters are not void _(although they are empty)_.  Sending `null`, as we currently do, has started to cause latest Rls code to crash parsing this notification.

### DIFF
--- a/lib/languageclient.js
+++ b/lib/languageclient.js
@@ -45,7 +45,7 @@ export class LanguageClientConnection {
 
   // Public: Send an `initialized` notification to the language server.
   initialized(): void {
-    this._sendNotification('initialized');
+    this._sendNotification('initialized', {});
   }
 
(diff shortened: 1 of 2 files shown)


alexheretic:
Thanks @daviwil, I'd also appreciate a release to avoid future Rls crashes. Any chance of getting `0.8.1` this week?

daviwil:
Sure!  I'll check with @damieng and see if there's anything else we'd want to get in for 0.8.1 this week.

damieng:
Thanks!  I think we just want the overridable method for determining the autocomplete auto trigger and we can put out 0.8.1

PR #172, opened 2018-01-18
Add language server restart API and auto-restart behavior

daviwil:
**Add AutoLanguageClient.restartServers API**

This change adds a new restartServers() API to AutoLanguageClient so
that implementors of the class can easily restart all active language
servers for their client in the workspace.

**Add language server auto-restart**

This change adds language server auto-restart behavior when a
language server exits or crashes unexpectedly.  The restart logic
allows a language server to be restarted up to 5 times within 3
minutes and then stops auto-restarting after reaching that limit.

### DIFF
--- a/lib/auto-languageclient.js
+++ b/lib/auto-languageclient.js
@@ -35,6 +35,7 @@ export default class AutoLanguageClient {
   _linterDelegate: linter$V2IndieDelegate;
   _signatureHelpRegistry: ?atomIde$SignatureHelpRegistry;
   _lastAutocompleteRequest: ?atom$AutocompleteRequest;
+  _isDeactivating: boolean;
 
   // Available if consumeBusySignal is setup
   busySignalService: ?atomIde$BusySignalService;
@@ -227,6 +228,7 @@ export default class AutoLanguageClient {
 
   // Deactivate disposes the resources we're using
   async deactivate(): Promise<any> {
+    this._isDeactivating = true;
     this._disposable.dispose();
     this._serverManager.stopListening();
     await this._serverManager.stopAllServers();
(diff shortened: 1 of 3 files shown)


lib/auto-languageclient.js
```diff
@@ -188,6 +194,14 @@ export default class AutoLanguageClient {
     return configuration;
   }
 
+  // Helper methods that are useful for implementors
+  // ---------------------------------------------------------------------------
+
+  // Restart all active language servers for this language client in the workspace
+  async restartServers() {
```
daviwil: Could also call this `restartAllServers`

lib/auto-languageclient.js
```diff
@@ -82,6 +82,12 @@ export default class AutoLanguageClient {
     return this.getGrammarScopes().includes(editor.getGrammar().scopeName);
   }
 
+  // Determine whether we should automatically restart a server that exited or
+  // crashed unexpectedly
+  shouldRestartFailedServer(server: ActiveServer): Promise<boolean> {
```
daviwil: Could also call this `shouldRestartExitedServer`

lib/auto-languageclient.js
```diff
@@ -82,6 +82,12 @@ export default class AutoLanguageClient {
...
+    return Promise.resolve(true);
```
daviwil: For now, default behavior is to always restart an exited server.  Should I add a simple check to see if the server has restarted more than X times in Y seconds and then return false if that threshold is reached?

damieng: Ideally yeah, I think the VSCode one restarts a maximum of 5 times in 3 minutes before reporting errors.  

We should also add some information about the number of restarts/times to the api if we go down that route.

I'd recommend we just auto-restart with the 5 times/3 minutes internally for now and open up the api in a future release if it's needed and we know more about what information is useful in determining whether to restart (e.g. cause, exit codes etc)

lib/server-manager.js
```diff
@@ -174,6 +175,13 @@ export class ServerManager {
     await Promise.all(this._activeServers.map(s => this.stopServer(s)));
   }
 
+  async restartAllServers(): Promise<void> {
+    this.stopListening();
+    await this.stopAllServers();
+    this._editorToServer = new Map();
```
damieng: We should probably check to make sure we're not leaking anything - e.g. around the document text sync listeners and the diagnostics.

daviwil: All listeners are registered with this._disposable and stopListening disposes of it, startListening creates a new one for new listeners.  I think we're covered on that front!

daviwil: I double-checked this and found that we're shutting down the "exclusive" adapters correctly in ServerManager, but I just converted the linting and signature help adapters to be added to the server's disposables list so that we can be a little more sure about when they are disposed just by looking at the code.  Let me know if that's undesirable and I can revert that commit!

j-f1:
Is there any way to make an error notification show up when you stop trying to restart the server?

damieng:
Looks great!  I do think we should tell the user if we give up trying to restart it though - at least for now.    should be as easy as dropping a call in to atom.notifications.error

daviwil:
Sounds good, will add that real quick!

daviwil:
Added this, how's the wording look?

image

damieng:
Looks good! Let's merge :)

daviwil:
Done, thanks a lot!

PR #175, opened 2018-02-07
Convert atom-languageclient to TypeScript

daviwil:
This PR is an experiment to see how much effort it'd be to convert the codebase to TypeScript.  Using TypeScript here would help us dogfood the ide-typescript package.

- [ ] Investigate ownership of official atom typings
- [ ] Investigate publishing atom-ide typings


lib/convert.ts
```diff
@@ -1,7 +1,6 @@
-// @flow
-
 import * as ls from './languageclient';
-import {Point, Range} from 'atom';
+import {Point, ProjectFileEvent, Range, TextEditor} from 'atom';
+import {Diagnostic, DiagnosticType, TextEdit} from '../typings/atom-ide';
```
daviwil: Ideally we wouldn't import from `../typings/atom-ide` but instead from a "fake"  `atom-ide` module provided by a `@types/atom-ide` package.

lib/convert.ts
```diff
@@ -191,7 +190,7 @@ export default class Convert {
       case 'Info':
         return ls.DiagnosticSeverity.Information;
       default:
-        (type: empty);
+        // (type: empty);
```
daviwil: This line gave an error in the TypeScript compiler, not sure what purpose it serves.

typings/atom-ide/index.d.ts
```diff
@@ -0,0 +1,267 @@
+import {Disposable, Grammar, Point, Range, TextEditor} from 'atom';
+
+/* declare module 'atom-ide' {*/
+
+  export type OutlineProvider = {
```
daviwil: This file is a conversion of the flowtype definitions for Atom-IDE and isn't finished yet.  There are some type system differences that I still need to account for.

lib/convert.ts
rictic: Assuming that the purpose was to assert that the switch was exhaustive over DiagnosticType, one way of writing this in typescript is:

```typescript
const never: never = type;
throw Error(`Unexpected diagnostic type ${never}`);
```

The `never` type is a type with no legal values, so that assignment will fail if type inference thinks that there are any legal values for `type` at that point in the program.

(Using `never` instead of `types` in the error ensures that it doesn't look like `never` is an unused variable when linting or compiling)

lib/convert.ts
```diff
@@ -202,7 +201,7 @@ export default class Convert {
   // * `textEdits` The language server protocol {TextEdit} objects to convert.
   //
   // Returns an {Array} of Atom {TextEdit} objects.
-  static convertLsTextEdits(textEdits: ?Array<ls.TextEdit>): Array<atomIde$TextEdit> {
+  static convertLsTextEdits(textEdits: Array<ls.TextEdit>): Array<TextEdit> {
```
rictic: The TypeScript representation of `?Array<ls.TextEdit>` is `Array<ls.TextEdit> | null | undefined` I think (unless you're not compiling with strict null checks)

tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "target": "es6",
+        "outDir": "build",
+        "lib": ["es6", "dom"],
+        "inlineSources": true,
+        "inlineSourceMap": true,
```
rictic: `"strict": true,` will catch many more errors, though it also has more false positives

daviwil: Thanks!  I think we'll turn on strictness once we make a bit more progress

lib/convert.ts
daviwil: Thanks for pointing that out!

daviwil:
Thanks for the help @rictic!

daviwil:
This PR is ready to be reviewed!  I'll sit down with @damieng when he's back and go over everything since this is a pretty big change.  Everyone else should feel free to ask questions!

damieng:
I'm going to have some time this morning so let's do it then!  I'll schedule with you on slack.

lib/auto-languageclient.ts
```diff
@@ -169,72 +196,72 @@ export default class AutoLanguageClient {
...
+  protected mapConfigurationObject(configuration: any): any {
     return configuration;
   }
 
   // Helper methods that are useful for implementors
   // ---------------------------------------------------------------------------
 
   // Gets a LanguageClientConnection for a given TextEditor
-  async getConnectionForEditor(editor: atom$TextEditor): Promise<?ls.LanguageClientConnection> {
+  protected async getConnectionForEditor(editor: TextEditor): Promise<ls.LanguageClientConnection> {
```
daviwil: Add `| null`

lib/auto-languageclient.ts
```diff
@@ -352,19 +381,23 @@ export default class AutoLanguageClient {
...
   // Start adapters that are not shared between servers
-  startExclusiveAdapters(server: ActiveServer): void {
+  private startExclusiveAdapters(server: ActiveServer): void {
     ApplyEditAdapter.attach(server.connection);
     NotificationsAdapter.attach(server.connection, this.name);
 
     if (DocumentSyncAdapter.canAdapt(server.capabilities)) {
-      server.docSyncAdapter = new DocumentSyncAdapter(server.connection, server.capabilities.textDocumentSync, editor =>
+      server.docSyncAdapter =
+        new DocumentSyncAdapter(server.connection, server.capabilities.textDocumentSync, (editor) =>
```
daviwil: Fix line breaking

lib/auto-languageclient.ts
```diff
@@ -545,15 +579,15 @@ export default class AutoLanguageClient {
...
+  public provideCodeFormat(): atomIde.RangeCodeFormatProvider {
     return {
       grammarScopes: this.getGrammarScopes(),
       priority: 1,
       formatCode: this.getCodeFormat.bind(this),
     };
   }
 
-  async getCodeFormat(editor: atom$TextEditor, range: atom$Range): Promise<Array<atomIde$TextEdit>> {
+  private async getCodeFormat(editor: TextEditor, range: Range): Promise<atomIde.TextEdit[]> {
```
daviwil: Make helper methods for providers protected?

lib/convert.ts
```diff
@@ -131,15 +131,15 @@ export default class Convert {
...
   // named equivalents.
-  static encodeHTMLAttribute(s: string): string {
+  public static encodeHTMLAttribute(s: string): string {
     const attributeMap = {
       '&': '&amp;',
       '<': '&lt;',
       '>': '&gt;',
       '"': '&quot;',
-      "'": '&apos;',
+      '\'': '&apos;',
```
daviwil: Disable this rule?

lib/download-file.ts
```diff
@@ -13,11 +11,11 @@ import fs from 'fs';
 //                      percentage on the callback can not be determined.
 //
 // Returns a {Promise} that will accept when complete.
-export default (async function downloadFile(
+export default (async function DownloadFile(
```
daviwil: Change this back to lower case

lib/languageclient.ts
```diff
@@ -0,0 +1,439 @@
...
+      this._log.debug(`rpc.sendRequest ${method} received (${Math.floor(took)}ms)`, result);
+      return result;
+    } catch (e) {
+      this._log.error(`rpc.sendRequest ${method} threw`, e);
+      throw e;
+    }
+  }
+}
+
+/**
```
daviwil: Move DiagnosticCode here

lib/logger.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export class ConsoleLogger {
+  public prefix: string;
+
+  constructor(prefix: string) {
+    this.prefix = prefix;
+  }
+
+  public warn(...args: any[]) {
+    // tslint:disable-next-line:no-console
```
daviwil: Can we disable this at file level?

lib/main.ts
```diff
@@ -1,11 +1,10 @@
...
 import LinterPushV2Adapter from './adapters/linter-push-v2-adapter';
 
-module.exports = {
-  AutoLanguageClient,
-  Convert,
-  DownloadFile,
-  LinterPushV2Adapter,
-};
+export * from './auto-languageclient';
+export { AutoLanguageClient, DownloadFile, Convert };
```
daviwil: Add Linter

package.json
```diff
@@ -5,50 +5,34 @@
...
-      "transpiler": "atom-babel6-transpiler",
-      "options": {
-        "cacheKeyFiles": [
-          "package.json",
-          ".babelrc"
-        ]
-      }
-    }
-  ],
+  "atomTestRunner": "./build/test/runner",
```
daviwil: Make sure this works in Atom test runner

lib/adapters/apply-edit-adapter.ts
```diff
@@ -49,7 +49,7 @@ export default class ApplyEditAdapter {
         const edits = Convert.convertLsTextEdits(changes[uri]);
         // Sort edits in reverse order to prevent edit conflicts.
         edits.sort((edit1, edit2) => -edit1.oldRange.compare(edit2.oldRange));
-        const buffer = editor.getBuffer();
+        const buffer = (editor as any).getBuffer() as TextBuffer;
```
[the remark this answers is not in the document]
daviwil: Turns out it is public and in the Atom type defs, fixable on our side

daviwil:
Updated the PR based on feedback and got tests running inside of Atom again.  Let me know if there's anything else you'd like me to do before merging!

lib/adapters/signature-help-adapter.ts
```diff
@@ -48,7 +47,7 @@ export default class SignatureHelpAdapter {
   }
 
   // Public: Retrieves signature help for a given editor and position.
-  getSignatureHelp(editor: atom$TextEditor, point: atom$Point): Promise<?atomIde$SignatureHelp> {
+  public getSignatureHelp(editor: TextEditor, point: Point): SignatureHelp | Promise<SignatureHelp> {
```
laughedelic: Shouldn't it be `Promise<lsp.SignatureHelp | null>`?

daviwil: Good catch, thanks!

laughedelic:
Could you clarify why all types `?Foo` became `Foo | null` and not `Foo | undefined | null` or `Foo | undefined`?

daviwil:
I suppose `Foo | undefined | null` would be more accurate.  I think I misread something in the TypeScript docs when I chose to do it that way.  I'll get this fixed soon!

laughedelic:
Cool! I just read about [`--strictNullChecks`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html) and it seem quite useful in this case.

daviwil:
Yep, I'm about 80% of the way to updating the codebase to be checked with that :)

damieng:
We only return null from these methods. I don't think we need undefined.

PR #177, opened 2018-02-14
Prevent showMessage spam + add projectPath to showMessage detail

alexheretic:
This pr improves _window/showMessage_ -> notification handling. 

Servers can resend the same error message, in this case it isn't much help to repeat the notifications.  We also currently lack an indication of which project the notification relates to.

### DIFF
--- a/lib/adapters/notifications-adapter.ts
+++ b/lib/adapters/notifications-adapter.ts
@@ -9,20 +9,22 @@ import {
   Notification,
   NotificationButton,
   NotificationOptions,
+  NotificationExt,
 } from 'atom';
 
 // Public: Adapts Atom's user notifications to those of the language server protocol.
 export default class NotificationsAdapter {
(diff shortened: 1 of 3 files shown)


laughedelic:
I made a hacky workaround for this problem: https://github.com/laughedelic/atom-ide-scala/pull/22, but it can deduplicate a notification only after it's already shown:



So this solution is so much better! 💯It addresses the problem at the right place.

damieng:
This looks great, thanks @alexheretic !

daviwil:
Hey @alexheretic, do you mind syncing latest master and reapply your changes there?  We've converted the codebase to TypeScript so you'll need to put these changes in the .ts version of the files you were editing.  I believe your changes should apply pretty cleanly.  Once that's done we can merge this in.  

Thanks a lot!

alexheretic:
No problem, I'll convert it tomorrow

daviwil:
Thanks a lot for putting up with that change :)  Merging now!

PR #179, opened 2018-02-15
Enable strictNullChecks in tsconfig.json

daviwil:
Please let me know if there are any places where these changes are incorrect!

/cc @laughedelic

### DIFF
--- a/lib/adapters/apply-edit-adapter.ts
+++ b/lib/adapters/apply-edit-adapter.ts
@@ -40,7 +40,7 @@ export default class ApplyEditAdapter {
       }),
     );
 
-    const checkpoints = [];
+    const checkpoints: Array<{ buffer: TextBuffer, checkpoint: number}> = [];
     try {
       for (let i = 0; i < editors.length; i++) {
         const editor = editors[i] as TextEditor;
@@ -52,7 +52,7 @@ export default class ApplyEditAdapter {
         const buffer = editor.getBuffer();
         const checkpoint = buffer.createCheckpoint();
         checkpoints.push({buffer, checkpoint});
-        let prevEdit = null;
+        let prevEdit: atomIde.TextEdit | null = null;
         for (const edit of edits) {
           ApplyEditAdapter.validateEdit(buffer, edit, prevEdit);
           buffer.setTextInRange(edit.oldRange, edit.newText);
@@ -74,9 +74,13 @@ export default class ApplyEditAdapter {
   }
 
   // Private: Do some basic sanity checking on the edit ranges.
-  private static validateEdit(buffer: TextBuffer, edit: atomIde.TextEdit, prevEdit: atomIde.TextEdit | null): void {
+  private static validateEdit(
+    buffer: TextBuffer,
+    edit: atomIde.TextEdit,
+    prevEdit: atomIde.TextEdit | null,
+  ): void {
     const path = buffer.getPath() || '';
-    if (prevEdit != null && edit.oldRange.end.compare(prevEdit.oldRange.start) > 0) {
+    if (prevEdit && edit.oldRange.end.compare(prevEdit.oldRange.start) > 0) {
       throw Error(`Found overlapping edit ranges in ${path}`);
     }
     const startRow = edit.oldRange.start.row;
--- a/lib/adapters/autocomplete-adapter.ts
+++ b/lib/adapters/autocomplete-adapter.ts
@@ -267,11 +267,11 @@ export default class AutocompleteAdapter {
   // * `editor` An Atom {TextEditor} used to obtain the necessary text replacement.
   // * `suggestion` An {atom$AutocompleteSuggestion} to set the replacementPrefix and text properties of.
   public static applyTextEditToSuggestion(
-    textEdit: TextEdit | null,
+    textEdit: TextEdit | undefined,
     editor: TextEditor,
     suggestion: AutocompleteSuggestion,
   ): void {
-    if (textEdit != null) {
+    if (textEdit) {
       suggestion.replacementPrefix = editor.getTextInBufferRange(Convert.lsRangeToAtomRange(textEdit.range));
       suggestion.text = textEdit.newText;
     }
@@ -296,7 +296,7 @@ export default class AutocompleteAdapter {
   //
   // Returns a {String} containing the AutoComplete+ suggestion type equivalent
   // to the given completion kind.
-  public static completionKindToSuggestionType(kind: number | null): string {
+  public static completionKindToSuggestionType(kind: number | undefined): string {
     switch (kind) {
       case CompletionItemKind.Constant:
         return 'constant';
--- a/lib/adapters/code-action-adapter.ts
+++ b/lib/adapters/code-action-adapter.ts
@@ -27,7 +27,7 @@ export default class CodeActionAdapter {
   public static async getCodeActions(
     connection: LanguageClientConnection,
     serverCapabilities: ServerCapabilities,
-    linterAdapter: LinterPushV2Adapter | null,
+    linterAdapter: LinterPushV2Adapter | undefined ,
     editor: TextEditor,
     range: Range,
     diagnostics: atomIde.Diagnostic[],
--- a/lib/adapters/datatip-adapter.ts
+++ b/lib/adapters/datatip-adapter.ts
@@ -88,5 +88,8 @@ export default class DatatipAdapter {
         value: markedString.value,
       };
     }
+
+    // Catch-all case
+    return { type: 'markdown', value: markedString.toString() };
   }
 }
--- a/lib/adapters/document-sync-adapter.ts
+++ b/lib/adapters/document-sync-adapter.ts
@@ -64,7 +64,7 @@ export default class DocumentSyncAdapter {
   //                    indicating whether this adapter should care about the contents of the editor.
   constructor(
     connection: LanguageClientConnection,
-    documentSyncKind: TextDocumentSyncOptions | number | null,
+    documentSyncKind: TextDocumentSyncOptions | number | undefined,
     editorSelector: (editor: TextEditor) => boolean,
   ) {
     this._connection = connection;
@@ -129,7 +129,7 @@ export default class DocumentSyncAdapter {
     );
   }
 
-  public getEditorSyncAdapter(editor: TextEditor): TextEditorSyncAdapter | null {
+  public getEditorSyncAdapter(editor: TextEditor): TextEditorSyncAdapter | undefined {
     return this._editors.get(editor);
   }
 }
--- a/lib/adapters/linter-push-v2-adapter.ts
+++ b/lib/adapters/linter-push-v2-adapter.ts
@@ -113,7 +113,7 @@ export default class LinterPushV2Adapter {
     if (path != null) {
       const diagnosticCodes = this._diagnosticCodes.get(path);
       if (diagnosticCodes != null) {
-        return diagnosticCodes.get(getCodeKey(range, text));
+        return diagnosticCodes.get(getCodeKey(range, text)) || null;
       }
     }
     return null;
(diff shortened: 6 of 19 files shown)


lib/adapters/apply-edit-adapter.ts
```diff
@@ -74,9 +74,13 @@ export default class ApplyEditAdapter {
   }
 
   // Private: Do some basic sanity checking on the edit ranges.
-  private static validateEdit(buffer: TextBuffer, edit: atomIde.TextEdit, prevEdit: atomIde.TextEdit | null): void {
+  private static validateEdit(
+    buffer: TextBuffer,
+    edit: atomIde.TextEdit,
+    prevEdit: atomIde.TextEdit | undefined | null,
```
daviwil: Could revert the addition of `undefined` here.

lib/adapters/autocomplete-adapter.ts
```diff
@@ -267,11 +267,11 @@ export default class AutocompleteAdapter {
   // * `editor` An Atom {TextEditor} used to obtain the necessary text replacement.
   // * `suggestion` An {atom$AutocompleteSuggestion} to set the replacementPrefix and text properties of.
   public static applyTextEditToSuggestion(
-    textEdit: TextEdit | null,
+    textEdit: TextEdit | undefined,
```
daviwil: This TextEdit comes from `CompletionItem.textEdit` which is typed as `TextEdit | undefined`, makes more sense to use `undefined` here

lib/adapters/autocomplete-adapter.ts
```diff
@@ -296,7 +296,7 @@ export default class AutocompleteAdapter {
   //
   // Returns a {String} containing the AutoComplete+ suggestion type equivalent
   // to the given completion kind.
-  public static completionKindToSuggestionType(kind: number | null): string {
+  public static completionKindToSuggestionType(kind: number | undefined): string {
```
daviwil: `CompletionItem.kind` uses `number | undefined` (or more accurately, a bunch of specific numbers unioned with `undefined`

lib/adapters/datatip-adapter.ts
```diff
@@ -88,5 +88,8 @@ export default class DatatipAdapter {
         value: markedString.value,
       };
     }
+
+    // Catch-all case
+    return { type: 'markdown', value: markedString.toString() };
```
daviwil: TypeScript compiler wouldn't let this method live without a catch-all case.  Is there a better result value?

lib/adapters/code-action-adapter.ts
```diff
@@ -27,7 +27,7 @@ export default class CodeActionAdapter {
   public static async getCodeActions(
     connection: LanguageClientConnection,
     serverCapabilities: ServerCapabilities,
-    linterAdapter: LinterPushV2Adapter | null,
+    linterAdapter: LinterPushV2Adapter | undefined ,
```
daviwil: `ActiveServer.linterPushV2` is typed as `LinterPushV2Adapter | undefined`, that gets passed into this method for this parameter

lib/adapters/document-sync-adapter.ts
```diff
@@ -64,7 +64,7 @@ export default class DocumentSyncAdapter {
   //                    indicating whether this adapter should care about the contents of the editor.
   constructor(
     connection: LanguageClientConnection,
-    documentSyncKind: TextDocumentSyncOptions | number | null,
+    documentSyncKind: TextDocumentSyncOptions | number | undefined,
```
daviwil: `ServerCapabilities.textDocumentSync` uses `undefined` instead of `null`

lib/adapters/linter-push-v2-adapter.ts
```diff
@@ -113,13 +113,13 @@ export default class LinterPushV2Adapter {
...
+        return diagnosticCodes.get(getCodeKey(range, text)) || null;
       }
     }
     return null;
   }
 }
 
 function getCodeKey(range: atom.Range, text: string): string {
-  return [].concat(...range.serialize(), text).join(',');
+  return ([] as any[]).concat(...range.serialize(), text).join(',');
```
daviwil: A little hack to make the type system happy

lib/adapters/linter-push-v2-adapter.ts
```diff
@@ -113,13 +113,13 @@ export default class LinterPushV2Adapter {
     if (path != null) {
       const diagnosticCodes = this._diagnosticCodes.get(path);
       if (diagnosticCodes != null) {
-        return diagnosticCodes.get(getCodeKey(range, text));
+        return diagnosticCodes.get(getCodeKey(range, text)) || null;
```
daviwil: Maps return `undefined` when a key has no value, so I'm coercing it to `null` to conform to the existing signature

lib/adapters/outline-view-adapter.ts
```diff
@@ -115,24 +115,27 @@ export default class OutlineViewAdapter {
...
     }
 
-    let parent = null;
+    let parent: atomIde.OutlineTree | undefined;
     for (const candidate of candidates) {
       if (
         candidate !== child &&
         candidate.startPosition.isLessThanOrEqual(child.startPosition) &&
-        (candidate.endPosition == null || candidate.endPosition.isGreaterThanOrEqual(child.endPosition))
+        (candidate.endPosition === undefined ||
```
daviwil: `endPosition` is an optional property which is `undefined` instead of `null`

lib/adapters/signature-help-adapter.ts
```diff
@@ -31,8 +31,8 @@ export default class SignatureHelpAdapter {
     const {signatureHelpProvider} = this._capabilities;
     assert(signatureHelpProvider != null);
 
-    let triggerCharacters = null;
-    if (Array.isArray(signatureHelpProvider.triggerCharacters)) {
+    let triggerCharacters: Set<string> | undefined;
```
daviwil: Not yet assigned, must use `undefined` in type

lib/auto-languageclient.ts
```diff
@@ -39,21 +39,6 @@ import SignatureHelpAdapter from './adapters/signature-help-adapter';
 
 export type ConnectionType = 'stdio' | 'socket' | 'ipc';
 
-// Public: Defines the minimum surface area for an object that resembles a
```
daviwil: Decided to move this to ServerManager's file because ServerManager was still using ChildProcess to refer to the language server process.

lib/convert.ts
```diff
@@ -159,7 +159,7 @@ export default class Convert {
       case 'deleted':
         return [{uri: Convert.pathToUri(fileEvent.path), type: ls.FileChangeType.Deleted}];
       case 'renamed': {
-        const results = [];
+        const results: Array<{ uri: string, type: ls.FileChangeType }> = [];
```
daviwil: TypeScript uses a type of `never[]` when you define a plain array with no typing.  You have to make the type specific so that the type system doesn't give a bunch of weird errors.

lib/download-file.ts
```diff
@@ -31,7 +31,7 @@ export default (async function downloadFile(
     throw Error('No response body');
   }
 
-  const finalLength = length || parseInt(response.headers.get('Content-Length' || '0'), 10);
+  const finalLength = length || parseInt(response.headers.get('Content-Length') || '0', 10);
```
daviwil: Bug in the previous code?

lib/download-file.ts
```diff
@@ -72,7 +72,7 @@ async function streamWithProgress(
       writer.write(Buffer.from(chunk));
       if (progressCallback != null) {
         bytesDone += chunk.byteLength;
-        const percent: number = length === 0 ? null : Math.floor(bytesDone / length * 100);
+        const percent: number | undefined = length === 0 ? undefined : Math.floor(bytesDone / length * 100);
```
daviwil: `percent` gets passed into ByteProgressCallback's optional `percent` parameter, `undefined` matches it

test/adapters/autocomplete-adapter.test.ts
```diff
@@ -20,7 +20,7 @@ describe('AutoCompleteAdapter', () => {
       capabilities: {completionProvider: { }},
       connection: new ls.LanguageClientConnection(createSpyConnection()),
       disposable: new CompositeDisposable(),
-      process: undefined,
+      process: undefined as any,
```
daviwil: Type system hack to not have to mock a whole process

test/adapters/code-highlight-adapter.test.ts
```diff
@@ -50,8 +50,10 @@ describe('CodeHighlightAdapter', () => {
       expect(highlightStub.called).to.be.true;
 
       invariant(result != null);
-      expect(result.length).to.equal(1);
-      expect(result[0].isEqual(new Range([0, 1], [0, 2]))).to.be.true;
+      if (result) {
```
daviwil: Tests required a bunch of null/undefined checks so that the compiler would be happy

typings/atom-ide/index.d.ts
```diff
@@ -178,7 +178,7 @@ export interface CodeActionProvider {
     editor: TextEditor,
     range: Range,
     diagnostics: Diagnostic[],
-  ): Promise<CodeAction[]>;
+  ): Promise<CodeAction[] | null>;
```
daviwil: The implementation of `getCodeActions` in `AutoLanguageClient` can return null, so had to add it here

laughedelic:
@daviwil thanks for keeping me in the loop. I should clarify that I'm neither Typescript, nor Flow expert 😅I'm writing a Scala.js type-facade for this library ([here](https://github.com/laughedelic/scalajs-atom-api/tree/master/src/main/scala/atom/languageclient)). So I just want to understand what's changing here, why and how it influences my work.

I'm going to go through all your comments, but before that have a general question: wouldn't it make more sense to use consistently _only one kind_ of "default"/absent value (`null` or `undefined`) everywhere in the code? (at least where you return them explicitly, I understand that external dependencies are out of control)

lib/adapters/linter-push-v2-adapter.ts
damieng: I wonder if we should just ``${range.serialize()},${text}`` here?

lib/download-file.ts
damieng: Ha, looks like it.  Nice catch.

lib/adapters/linter-push-v2-adapter.ts
daviwil: I'll give it a shot!

daviwil:
@laughedelic It seems that there's [a lot of debate](https://github.com/Microsoft/TypeScript/issues/7426) over `null` vs `undefined` and whether `null` should ever be used.  In my changes here, I decided to keep the use of `null` where it was intended to mean "no result" so that we don't break existing packages and then change any other usages when a value or property hadn't been defined.  In the cases where we definitely return null, I didn't add `| undefined` to the type just so that the return type is clear.

Let me know if you have any other questions!

lib/adapters/linter-push-v2-adapter.ts
daviwil: `range.serialize()` won't work because it returns an array of arrays, that's why the `[].concat()` with the spread operator is used.  It builds a flat array of items that can be joined.  I'll leave it as it is for now

laughedelic:
@daviwil thanks for the answer and the link! That thread is huge, but has a lot of interesting opinions. I got your point, it will be useful for me when working with the library. The inconvenience of using `null` in Scala.js is that you can't enforce `strictNullChecks` to make `null` a special value which doesn't belong to any other type, in Scala it's a value of any type, so you cannot really say `T | null` explicitly, because it's the same as just `T`, while `undefined` _is_ a special value (the only value of it's type) and it can be explicitly separated in the type annotation `T | undefined` (`UndefOr[T]` in Scala.js and is similar to the common `Option[T]`).

PR #180, opened 2018-02-17
Fix type exports for atom-ide and atom-linter.

GlenCFL:
I am trying to use the types exported by this package a bit early, as I want to rewrite a package of mine using them once they're published to NPM. I've run into an issue regarding the export of the two modules within the `typings` directory.

These modules aren't accessible to the definitions within `build/lib` once installed, as the `tsconfig.json` file isn't used by those importing. The easiest way I know to solve this is to just reference them with the main source file of the library and use the typical module syntax for the modules themselves, which I've done with this pull request. This doesn't force users to use the reference imports themselves or have the path values in their `tsconfig.json` file.

### DIFF
--- a/lib/adapters/linter-push-v2-adapter.ts
+++ b/lib/adapters/linter-push-v2-adapter.ts
@@ -12,9 +12,9 @@ import Convert from '../convert';
 // Public: Listen to diagnostics messages from the language server and publish them
 // to the user by way of the Linter Push (Indie) v2 API supported by Atom IDE UI.
 export default class LinterPushV2Adapter {
-  private _diagnosticMap: Map<string, linter.V2Message[]> = new Map();
+  private _diagnosticMap: Map<string, linter.Message[]> = new Map();
   private _diagnosticCodes: Map<string, Map<string, DiagnosticCode | null>> = new Map();
-  private _indies: Set<linter.V2IndieDelegate> = new Set();
+  private _indies: Set<linter.IndieDelegate> = new Set();
 
   // Public: Create a new {LinterPushV2Adapter} that will listen for diagnostics
   // via the supplied {LanguageClientConnection}.
@@ -32,7 +32,7 @@ export default class LinterPushV2Adapter {
   // Public: Attach this {LinterPushV2Adapter} to a given {V2IndieDelegate} registry.
   //
   // * `indie` A {V2IndieDelegate} that wants to receive messages.
-  public attach(indie: linter.V2IndieDelegate): void {
+  public attach(indie: linter.IndieDelegate): void {
     this._indies.add(indie);
     this._diagnosticMap.forEach((value, key) => indie.setMessages(key, value));
     indie.onDidDestroy(() => {
@@ -73,7 +73,7 @@ export default class LinterPushV2Adapter {
   // * `diagnostics` A {Diagnostic} object received from the language server.
   //
   // Returns a {V2Message} equivalent to the {Diagnostic} object supplied by the language server.
-  public diagnosticToV2Message(path: string, diagnostic: Diagnostic): linter.V2Message {
+  public diagnosticToV2Message(path: string, diagnostic: Diagnostic): linter.Message {
     return {
       location: {
         file: path,
--- a/lib/auto-languageclient.ts
+++ b/lib/auto-languageclient.ts
@@ -47,7 +47,7 @@ export type ConnectionType = 'stdio' | 'socket' | 'ipc';
 export default class AutoLanguageClient {
   private _disposable = new CompositeDisposable();
   private _serverManager: ServerManager;
-  private _linterDelegate: linter.V2IndieDelegate;
+  private _linterDelegate: linter.IndieDelegate;
   private _signatureHelpRegistry: atomIde.SignatureHelpRegistry | null;
   private _lastAutocompleteRequest: AutocompleteRequest;
   private _isDeactivating: boolean;
@@ -514,7 +514,7 @@ export default class AutoLanguageClient {
   }
 
   // Linter push v2 API via LS publishDiagnostics
-  public consumeLinterV2(registerIndie: (params: {name: string}) => linter.V2IndieDelegate): void {
+  public consumeLinterV2(registerIndie: (params: {name: string}) => linter.IndieDelegate): void {
     this._linterDelegate = registerIndie({name: this.name});
     if (this._linterDelegate == null) {
       return;
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -1,5 +1,8 @@
-// tslint:disable-next-line:no-reference
+// tslint:disable:no-reference
 /// <reference path="../typings/atom/index.d.ts"/>
+/// <reference path="../typings/atom-ide/index.d.ts"/>
+/// <reference path="../typings/atom-linter/index.d.ts"/>
+// tslint:enable:no-reference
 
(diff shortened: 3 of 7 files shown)


daviwil:
Hey @GlenCFL, thanks for sending this! The reason we didn't export the types in this way is because we are considering shipping them via DefinitelyTyped. I had assumed that it wouldn't be necessary to access them just yet; can you give me an example of a case where you need access to these types to use atom-languageclient?

Thanks!

GlenCFL:
Having the `types` field within the `package.json` file means that the types are automatically pulled in whenever you install the `atom-languageclient` package. If the definition file that is pulled in contains any errors, then you cant even import without your TypeScript build failing.

If you want to publish to DefinitelyTyped, then you'd remove that. It's preferred for you to bundle your types with your package, as mentioned in the [TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html), but I'd assume there is a reason you're leaning in that direction.

daviwil:
Interesting, I've been using this module in another package without any issues so far, but I've got it npm linked, so I don't know if that makes a difference. I'll look into this a bit more next week and see if we need to include/export the types in this way.

Thanks again!

daviwil:
Hey @GlenCFL, since you're one of the maintainers of Atom's DefinitelyTyped definitions, I'd like to get your opinion on what path we should take:

1. Ship Atom IDE typings via DefinitelyTyped so that they can be used/updated independently of atom-languageclient (is there any value in that?)

2. Include Atom IDE typings with the atom-languageclient package.  I like how clean your solution is here but I want to make sure there's no advantage to shipping them separately.

Also, should we try to merge the atom-linter types here back into the [ones in the DT atom typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/atom/linter/index.d.ts)?

It'd be great to get these things figured out before the first release containing the new TypeScript typings so that it won't require additional changes to consumers' package.json files if we change course.

@damieng Thoughts?

GlenCFL:
The directories specific to packages on DefinitelyTyped tend to be reserved only for those that are published to NPM, which doesn't really fit AtomIDE. Obviously there are exceptions made, as otherwise `atom` itself wouldn't exist, but those types date back to a time before Microsoft became as involved as they are now. I don't know if there is much value to publishing separately, as AtomIDE and this package tend to be advertised hand-in-hand. Maybe @lierdakil has more insight as a package author supporting multiple languages.

The types could always be published to a subdirectory of `atom` as well, though those directories are not versioned in the same way. If you were to just shorthand it to 'ide', then imports would be:

```ts
import { ... } from "atom/ide"
```

One thing to note if you go the separate package route is that you should put several authors onto the definition file. Reviews are the way to get on the quick merge path on DefinitelyTyped. If you do go with a subdirectory, then you're welcome to add yourselves to `types/atom`.

As for the Linter definitions, I'll look over them in a bit more depth. At first glance, the big difference is that you want to support both V1 and V2, whereas V1 has been dropped within our file. V1 could be added back, but the names would be different in order to prevent breakage.

daviwil:
I'm inclined to agree that atom-languageclient and Atom IDE's TypeScript typings go hand in hand and maybe should be shipped together in the atom-languageclient module.  It seems there could be a clear migration path if we did decide to separate them in the future, just add `@types/atom[-ide]` to the `dependencies` of atom-languageclient so that it gets pulled in for consumers, so maybe we don't need to fully decide right now.

Would certainly appreciate you taking a look at whether it makes sense to merge the Linter typings but maybe it isn't extremely critical for those to come from the same place.

GlenCFL:
Quick question. Is this type wrong or do you store internally as Linter V1 messages?

```ts
export interface V2IndieDelegate {
    name(): string;
    getMessages(): Message[]; // Wrong message type?
    clearMessages(): void;
    setMessages(filePath: string, messages: V2Message[]): void;
    setAllMessages(messages: V2Message[]): void;
    onDidUpdate(callback: () => void): Disposable;
    onDidDestroy(callback: () => void): Disposable;
    dispose(): void;
}
```

Aside from this one property, all V2 fields are identical.

damieng:
You can use atom-ide-ui without atom-languageclient so I'm not convinced we need to force them together.

lierdakil:
I'm with @damieng on this. atom-languageclient is a good way to get the thing working quickly, but it's not the *only* way to get atom-ide support.

Case in point: atom-typescript package, which uses tsserver directly, can use atom-ide-ui. But it can't use atom-languageclient in any meaningful way.

So my vote goes to putting atom-ide-ui types on DT.

daviwil:
@lierdakil any preference on the Atom IDE types being delivered as a subfolder of the DT Atom typings or in their own separate 'atom-ide' folder/module?

lierdakil:
@daviwil, no preference really, either could work. But don't I think it would make much *sense* outside of atom subfolder. Since, you know, you need Atom to have atom-ide-ui, it's not like it's an NPM module. And since DT ships linter types with 'atom', I see no reason why not ship 'atom-ide-ui' types as well.

GlenCFL:
The `linter-legacy` subdirectory is being added to `types/atom`, which will have the missing types necessary to get rid of the `atom-linter` module. Your Atom augmentations are already referenced, which just leaves the AtomIDE module.

hansonw:
We'd be open to maintaining a `@types/atom/atom-ide-ui` if that's what the community wants. (I assume some kind of Flow-Typescript converter already exists somewhere so it seems automatable? Not certain about that though.)

damieng:
I don't think automation works very well - there are lots of gotcha's to be considered.

We manually converted it here which we could check in to atom-ide-ui to maintain there alongside the flow types?  https://github.com/atom/atom-languageclient/blob/master/typings/atom-ide/index.d.ts

hansonw:
Does it need to be checked into atom-ide-ui? It seems like the other Atom definitions are source-of-truthed on DefinitelyTyped right?

daviwil:
Yep, better to have then in the DT repo otherwise it'll be a pain keeping them in sync.

GlenCFL:
An update to `vscode-languageserver-protocol` has broken the builds. 3.6 might be published soon.

I was going to remove the V1 types and fix the above interface, then recommend that you just keep them in repository for now. The bot on DefinitelyTyped has been broken this entire week, so nothing is being merged there.

GlenCFL:
The latest commit simply pulls in the `atom/linter` types, renames symbols to match the new names, and temporarily locks the LSP version until someone else wants to look into the failure.

Switching over to `atom/linter` types later on would be trivial, but again nothing is being merged there.

daviwil:
We also noticed the break from the package updates this morning, thanks a lot for fixing that!  I like where the PR is currently at, and definitely agree with making the atom/linter types match what you updated in DefinitelyTyped.

@damieng, do you have any feedback on the changes to the Linter types?  Ultimately we'll be using the updated Linter types in DT that look just like this.

damieng:
This all looks good to me. Am also running into the build breaks.  If these linter types match those on DT should we just start using them directly from there rather than having our own version here?

GlenCFL:
The DT definitions are missing one property, `linterName`, which you currently use within your code. Both the "atom-linter" module and the reference to it can be removed once [DefinitelyTyped#23804](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23804) is merged, but that will likely be a week from now due to their bot being missing.

damieng:
Thanks for all your help on this @GlenCFL !

PR #181, opened 2018-02-17
 Enable noImplicitAny within the tsconfig.

GlenCFL:
Aside from #180, there were also a lot of `noImplicitAny` errors within the definitions, primarily due to mislabeled parameters. I've enabled the rule within the `tsconfig.json` file and fixed the resulting errors within this pull request.

### DIFF
--- a/lib/server-manager.ts
+++ b/lib/server-manager.ts
@@ -16,13 +16,15 @@ import { CompositeDisposable, ProjectFileEvent, TextEditor } from 'atom';
 // ChildProcess.  This is used so that language packages with alternative
 // language server process hosting strategies can return something compatible
 // with AutoLanguageClient.startServerProcess.
-export interface LanguageServerProcess extends cp.ChildProcess {
+export interface LanguageServerProcess extends EventEmitter {
   stdin: stream.Writable;
   stdout: stream.Readable;
   stderr: stream.Readable;
   pid: number;
 
   kill(signal?: string): void;
+  on(event: 'error', listener: (err: Error) => void): this;
+  on(event: 'exit', listener: (code: number, signal: string) => void): this;
 }
 
 // The necessary elements for a server that has started or is starting.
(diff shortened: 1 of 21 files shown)


damieng:
This looks great, thanks!  Look okay to you too @daviwil ?

lib/server-manager.ts
```diff
@@ -16,15 +16,13 @@ import { CompositeDisposable, ProjectFileEvent, TextEditor } from 'atom';
 // ChildProcess.  This is used so that language packages with alternative
 // language server process hosting strategies can return something compatible
 // with AutoLanguageClient.startServerProcess.
-export interface LanguageServerProcess extends EventEmitter {
+export interface LanguageServerProcess extends cp.ChildProcess {
```
daviwil: We definitely don't want to extend from ChildProcess here because we're trying to create an interface that exposes the minimal ChildProcess interface that we use in atom-languageclient (see the comment above the interface).  Was there some particular error you made this change to address?

More background on this: in the [ide-powershell](https://github.com/daviwil/ide-powershell) package, we use node-pty to spawn the language server process and the object it returns does not conform to the ChildProcess API shape.  I have to make a wrapper object that exposes the minimum set of methods and events that atom-languageclient uses so that the PowerShell language server can be returned from startServerProcess.  If this LanguageServerProcess interface changes to inherit from ChildProcess, it will require me to fake the whole interface even though we don't use anything else on it.

Can you think of another way to solve this issue?

GlenCFL: The only real issue from the type definition standpoint is that this type is eventually fed into the RPC `IPCMessageReader` and `IPCMessageWriter`. If you are sure that this will work, then we could always just perform a TypeScript cast to `ChildProcess` at those two call sites.

daviwil: Argh, yeah, that's annoying.  However it seems that their use of ChildProcess is pretty minimal as well (moreso than LanguageServerProcess) so I'd say let's cast to ChildProcess for now and then see if any future problems arise.

daviwil:
Thanks a lot for making that change and for sending this PR!

PR #184, opened 2018-02-26
Log request cancellation errors at debug level

daviwil:
Gets rid of some error noise in log output caused by uncaught promise rejections when CancellationTokens are cancelled.


daviwil:
PR updated to shift request cancellation error logging to debug level.

PR #185, opened 2018-02-27
Autocomplete improvements

damieng:
This reworks autocompletion (again) to:

- Trigger on trigger character OR a min char count specified in autocomplete-plus settings (defaults to 3)
- Always filter the results in case the server does not look back from the cursor position (some do not)
- Simplify the code

Note: Some static methods marked *public* have been changed or removed. We'll use semver once we hit 1.0.0.

### DIFF
--- a/lib/adapters/autocomplete-adapter.ts
+++ b/lib/adapters/autocomplete-adapter.ts
@@ -68,9 +68,9 @@ export default class AutocompleteAdapter {
     const triggerColumn = request.bufferPosition.column - prefixWithTrigger.length;
     const triggerPoint = new Point(request.bufferPosition.row, triggerColumn);
 
-    // We auto-trigger on a trigger character or after the minimum number of characters from autocomplete-plus
+    // Only auto-trigger on a trigger character or after the minimum number of characters from autocomplete-plus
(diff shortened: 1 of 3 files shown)


lib/adapters/autocomplete-adapter.ts
```diff
@@ -62,31 +63,42 @@ export default class AutocompleteAdapter {
...
         server.capabilities.completionProvider.triggerCharacters || [] : [];
-    const triggerPoint = AutocompleteAdapter.getTriggerPoint(request, triggerChars);
-    const prefixWithTrigger = AutocompleteAdapter.getPrefixWithTrigger(request, triggerPoint);
+    const triggerChar = AutocompleteAdapter.getTriggerCharacter(request, triggerChars);
+    const prefixWithTrigger = triggerChar + request.prefix;
+    const triggerColumn = request.bufferPosition.column - prefixWithTrigger.length;
+    const triggerPoint = new Point(request.bufferPosition.row, triggerColumn);
+
+    // Only auto-trigger on a trigger character or after the minimum number of characters from autocomplete-plus
+    const minimumWordLength = atom.config.get('autocomplete-plus.minimumWordLength') || 0;
```
daviwil: This couples atom-languageclient to autocomplete-plus, is there a way we can have the client register a method we can call to get the current word length?

PR #186, opened 2018-02-27
Remove the atom-linter module.

GlenCFL:
This is a quick followup to #180. The DefinitelyTyped PR was finally merged, so the `atom-linter` module can now be removed.


daviwil:
Thanks for taking this all the way @GlenCFL!

PR #188, opened 2018-03-07
Add getters to Autocomplete suggestions priority

martinfojtik:
This can be useful when you want to change autocomplete suggestions priority inside LSP package.

### DIFF
--- a/lib/auto-languageclient.ts
+++ b/lib/auto-languageclient.ts
@@ -86,6 +86,18 @@ export default class AutoLanguageClient {
     throw Error('Must implement getServerName when extending AutoLanguageClient');
   }
 
+  protected getInclusionPriority(): number {
+    return 1;
+  }
+
+  protected getSuggestionPriority(): number {
+    return 2;
+  }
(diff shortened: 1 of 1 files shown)


damieng:
I think we'd either want them to override `provideAutocomplete` or wire these up to read config settings automatically as AutoLanguageClient's surface area is already too big.  Thoughts?

martinfojtik:
I think overriding `provideAutocomplete` is riskier than overriding this new getters. Because if in future we make new changes inside `provideAutocomplete`, it could introduce breaking change, and we will be forced to propagate that code change into other plugins. @damieng What do you think?

damieng:
The easiest option I think is to call down to the base provideAutocomplete and then change the properties, e.g.

```
provideAutocomplete() {
  const provided = super.provideAutocomplete()
  provided.inclusionPriority = ... // get config
  provided.suggestionPriority = ... // get config
  provided.excludeLowerPriority = ... // get config
  return provided
}
```

Issue #155, opened 2017-11-27
Add support for dynamic capability registration


daviwil:
Language Server Protocol v3 introduces dynamic registration of capabilities, allowing the server to register support of a particular capability with the client, also passing options to configure the client's use of that capability.

This will require us to use a different set of method overloads in vscode-jsonrpc's client library for sending requests and events:

https://github.com/Microsoft/vscode-languageserver-node/blob/master/jsonrpc/src/main.ts#L269

daviwil:
Actually, it seems the RegistrationOptions for the given Request/Notification types aren't even being used in those overloads, they're just a facet of the Request/NotificationType signature.  My statement about changing the overloads was incorrect.  The largest part of the work will be in dealing with the `client/registerCapability` request to dynamically wire up language features.

The real reason we would need to change overloads is to make use of the CancellationToken parameter which has nothing to do with dynamic capability registration.  Maybe in the future we can make a PR to the vscode-jsonrpc project to add overloads that take a CancellationToken without the need for a Request/NotificationType parameter.

By the way, I love your avatar, that movie is awesome ;)

deankevorkian:
So, for now we're only going to cancel requests via a `$/cancelRequest` request until we make a PR to vscode-jsonrpc or if Request/Notification types start receiving a `CancellationToken` object? I like it. 
I understand it's a part of vscode anyway, so why hasn't it been an existing overload on `vscode-languageserver-node` this whole time? :confused: 

> By the way, I love your avatar, that movie is awesome ;)
;)

daviwil:
Probably wasn't added as an overload to the "simple" sendRequest/sendNotification methods because vscode-languageclient switched over to the new method of specifying request and notification types before adding request cancellation.  I doubt there's a limitation that prevents it from being added, so I'll try to add it and send a PR to them.

daviwil:
Turns out vscode-jsonrpc does support a CancellationToken parameter in the "overload" of sendRequest which accepts a simple method name and parameter list:

https://github.com/Microsoft/vscode-languageserver-node/blob/master/jsonrpc/src/main.ts#L928

They just look at the last item in the list of params and check if it's a CancellationToken.  I'll add this information to issue #140 as well.

UziTech:
Development of [atom-languageclient](https://www.npmjs.com/package/atom-languageclient) has officially moved to https://github.com/atom-ide-community/atom-languageclient 🎉 

If this is still an issue please consider opening an issue on that repo.

PR #202, opened 2018-03-24
Respect server document sync capabilities

laughedelic:
See https://github.com/atom/atom-languageclient/pull/193#discussion_r176617062 for the context. This is a minor improvement avoiding breaking changes:
* don't send `didOpen`/`didClose` notifications if `documentSync.openClose` is _explicitly disabled_
* add document content if the `documentSync.save.includeText` is enabled

----

And here are some thoughts for discussion.

I think that current behaviour is not completely correct, but fixing it may break somebody's server. IMO in LSP v3 all capability options have to be enabled explicitly. So if the server wants to receive open/close notifications, it should set `openClose: true`, same as `willSave`/`willSaveWaitUntil`. 

* right now these options are treated differently: `willSave*` has to be enabled explicitly, but `openClose` and `save` not.
* with current behaviour one still can disable `open`/`close`/`willSave*` notifications, but cannot disable `save` (because it's not a boolean and undefined  is implicitly treated as `true`)
* `change` option is treated specially: it determines whether all other notifications are even considered

An example: if a server cares only about `open`, `close` and `willSaveWaitUntil` notifications, but doesn't want to receive any other, it cannot communicate it. Turning off `change` notifications while receiving some others can be useful if server has its own file watching mechanism. But currently it will result in `DocumentSyncAdapter.canAdaptV3` returning `false` and not receiving any notifications at all.

Also, I'm not sure about this, but I think that [`didChangeWatchedFiles`](https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md#didchangewatchedfiles-notification-arrow_right) notifications have to be registered independently from all above.

### DIFF
--- a/lib/adapters/document-sync-adapter.ts
+++ b/lib/adapters/document-sync-adapter.ts
@@ -143,6 +143,7 @@ export class TextEditorSyncAdapter {
   private _connection: LanguageClientConnection;
   private _fakeDidChangeWatchedFiles: boolean;
   private _versions: Map<string, number>;
+  private _documentSync: TextDocumentSyncOptions;
 
   // Public: Create a {TextEditorSyncAdapter} in sync with a given language server.
   //
@@ -159,6 +160,7 @@ export class TextEditorSyncAdapter {
     this._connection = connection;
     this._versions = versions;
     this._fakeDidChangeWatchedFiles = atom.project.onDidChangeFiles == null;
+    this._documentSync = documentSync;
 
     const changeTracking = this.setupChangeTracking(documentSync);
     if (changeTracking != null) {
@@ -172,14 +174,20 @@ export class TextEditorSyncAdapter {
     if (documentSync.willSaveWaitUntil) {
       this._disposable.add(editor.getBuffer().onWillSave(this.willSaveWaitUntil.bind(this)));
     }
+    // Send close notifications unless it's explicitly disabled
+    if (documentSync.openClose !== false) {
+      this._disposable.add(editor.onDidDestroy(this.didClose.bind(this)));
+    }
     this._disposable.add(
       editor.onDidSave(this.didSave.bind(this)),
-      editor.onDidDestroy(this.didClose.bind(this)),
       editor.onDidChangePath(this.didRename.bind(this)),
     );
 
     this._currentUri = this.getEditorUri();
-    this.didOpen();
+
+    if (documentSync.openClose !== false) {
+      this.didOpen();
+    }
   }
 
   // The change tracking disposable listener that will ensure that changes are sent to the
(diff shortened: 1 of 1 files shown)


damieng:
>I think that current behaviour is not completely correct, but fixing it may break somebody's server. IMO in LSP v3 all capability options have to be enabled explicitly. So if the server wants to receive open/close notifications, it should set openClose: true, same as willSave/willSaveWaitUntil.

The problem is that openClose capabilities flag does not exist in LSP v2 but didSave and didOpen messages do so back in v2 you definitely had to always send it to be compliant.  There is no version level negotiation between a client and server and we definitely still want to support v2 servers so I think what you've done here - defaulting to sending it unless specified otherwise - is the right thing to do.

Save and open events are quite lightweight so I don't think it's too bad for a server to just ignore what they don't want.

> Also, I'm not sure about this, but I think that didChangeWatchedFiles notifications have to be registered independently from all above.

I thought that one was capabilities flagged but it seems that was missed.

laughedelic:
> The problem is that openClose capabilities flag does not exist in LSP v2 but didSave and didOpen messages do so back in v2 you definitely had to always send it to be compliant. 

OK, I see that the current behaviour is justified by the compatibility with LSP v2. 

> There is no version level negotiation between a client and server

But in this particular context there is clear distinction between v2 and v3: the type of the sync options is either `TextDocumentSyncKind`/`number` or `TextDocumentSyncOptions`. We could preserve compatibility with the LSP v2 servers by setting those defaults explicitly when converting from `number` to `TextDocumentSyncOptions`:

```js
  this._documentSync = {
    openClose: true,
    change: documentSync || TextDocumentSyncKind.Full,
    save: { includeText: false }, // not sure what was the default in v2
    // willSave options were not present in v2 AFAIK
  };
```

While if the server sends `TextDocumentSyncOptions` it is treated as is and absent options are interpreted as `false`. What do you think of this approach?

> Save and open events are quite lightweight so I don't think it's too bad for a server to just ignore what they don't want.

I agree that it's not a big deal, but if it's easy to improve, why not 😊  

> > Also, I'm not sure about this, but I think that didChangeWatchedFiles notifications have to be registered independently from all above.
> 
> I thought that one was capabilities flagged but it seems that was missed.

I guess, fixing this will require implementing #155 first.

damieng:
That logic sounds good to me but like you I could go either way so it's up to you :)

laughedelic:
Let's leave it as is. Now there is  an option to disable those features (except `didSave`), so if server wants or doesn't want some notifications, it can be explicit about it one way or the other, so it doesn't matter how undefined values are interpreted implicitly.

laughedelic:
Thanks!

Issue #200, opened 2018-03-23
Set a timeout on willSaveWaitUntil request


laughedelic:
See https://github.com/atom/atom-languageclient/pull/193/files/5a9894f84aae9542fc710ca54b6a7e14142fea04#r176509321

From the [LSP spec](https://microsoft.github.io/language-server-protocol/specification#textDocument_willSaveWaitUntil):

> Please note that clients might drop results if computing the text edits took too long or if a server constantly fails on this request. This is done to keep the save fast and reliable.

damieng:
Do you want to have a crack at this before we put a new version out or just go with a new version?

laughedelic:
I had some code for this, but I wasn't sure I'm doing it right, so I just left it out of #193. I'll just open a PR and you tell me if it's worth working on or better to leave it for later (for somebody more prepared).

laughedelic:
I see that you already tagged v0.9.5 👍 So I can already start using it. I think it's better to have frequent releases and this particular thing is not critical to block it.

laughedelic:
Fixed in #203

PR #203, opened 2018-03-30
Set a timeout on the willSaveWaitUntil handler

laughedelic:
Addresses #200. I have several doubts here:

1. I added this `Utils.promiseWithTimeout` combined from several solutions found on StackOverflow. It seems there is no canonical way of doing this, so I'm not sure this is correct.

1. I randomly set a hardcoded 2.5 seconds timeout. I've actually seen server replying to this request with formatting edits in 3.5s on a big file. It's hard to choose a reasonable time limit: user may get frustrated with the delay on save as well as with the constant error messages. Probably a higher time limit + a spinning wheel while would be a good compromise?

1. I think it would make sense to cancel request on timeout, but I don't know how to do it: `LanguageClientConnection` doesn't have a method for `$/cancelRequest` and even if it did, where do I get the request id? I see `cancellationToken`s somewhere, but is it only for the server-cancelled requests?

### DIFF
--- a/lib/adapters/document-sync-adapter.ts
+++ b/lib/adapters/document-sync-adapter.ts
@@ -20,7 +20,6 @@ import {
 } from 'atom';
 import Utils from '../utils';
 import { BusySignalService } from 'atom-ide';
-import * as path from 'path';
 
 // Public: Synchronizes the documents between Atom and the language server by notifying
 // each end of changes, opening, closing and other events as well as sending and applying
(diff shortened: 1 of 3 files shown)


laughedelic:
I went ahead and added busy-signal indicator. I think it's much better now: if it's fast, it's just like a normal save, if it's slow, you see an indicator, and if it's too slow, you will get a timeout error notification (btw, messages text could be improved).

I also added a little improvement: trying to put the cursor where it was before formatting. Currently if the server sends a full buffer replace `TextEdit`, after being applied, cursor jumps to the end. This is very annoying. My fix in 6cd5654 is just a little improvement, as there are many more nuances: handling multiple cursors, attaching to some syntax node position could be much better, preserving folds could be nice if it's possible. This definitely deserves a separate issue.

PR #206, opened 2018-04-13
Split off AtomIDE-specific parts from AutoLanguageClient

lierdakil:
I was doing something strange with the code, namely, using it to simplify working with a language server but not actually using any adapters (because I was trying to adapt it to a completely different stack), and the code was getting increasingly monkey-patchey, to the point of hurting my brain (and pride).

So here I come with a proposal: can we split parts that are not directly dependent on adapters into `BaseLanguageClient` and export that?

I also enabled TypeScripts's strict mode and some optional checks, added some types and fixed a bug or two while at it.

There are more things on top of this that I did, namely cleaning up Atom typings and adding a prettier code formatter to the set-up (which I find saves a lot of keystrokes when tslint is set up to complain about formatting), but decided to split those out for the sake of simplicity. When (if) this is merged, I will create more PRs.

The main idea here is to untie adapters from `ActiveServer`, `ServerManager` and all `AutoLanguageClient` parts that don't depend on adapters directly (the latter are moved more or less verbatim to `BaseLanguageClient`). The only tricky parts are in `server-manager.ts`, specifically, `unsupportedEditorGrammar` is a bit of a hack (but let's face it, that part was a bit of a hack in the first place)

As a side effect, *some* properties that were marked `private` now became `protected`. I don't think this is a huge deal since there's no concept of `private` in JS in the first place, but let me know if you think otherwise.

There's one bugfix related to incorrect type for `TextEditor.getNonWordCharacters` -- just something I noticed in passing. Feel free to cherry-pick.

There's another type-related fix, where `scopeDescriptor` type in autocomplete-plus completion request isn't right, but it doesn't seem like it affects the actual code.

### DIFF
--- a/lib/adapters/apply-edit-adapter.ts
+++ b/lib/adapters/apply-edit-adapter.ts
@@ -29,7 +29,7 @@ export default class ApplyEditAdapter {
     try {
       // Sort edits in reverse order to prevent edit conflicts.
       edits.sort((edit1, edit2) => -edit1.oldRange.compare(edit2.oldRange));
-      edits.reduce((previous, current) => {
+      edits.reduce((previous: atomIde.TextEdit | null, current) => {
         ApplyEditAdapter.validateEdit(buffer, current, previous);
         buffer.setTextInRange(current.oldRange, current.newText);
         return current;
--- a/lib/adapters/document-sync-adapter.ts
+++ b/lib/adapters/document-sync-adapter.ts
@@ -96,7 +96,7 @@ export default class DocumentSyncAdapter {
   //
   // * `editor` A {TextEditor} to consider for observation.
   public observeTextEditor(editor: TextEditor): void {
-    const listener = editor.observeGrammar((grammar) => this._handleGrammarChange(editor));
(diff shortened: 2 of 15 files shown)


lierdakil:
Ping. Any comment at all?

damieng:
Sorry, this is a significant change that also somewhat changes the direction of the project which was always to be a bridge between the LSP and the Atom IDE UI services.

Some of the refactoring and clean-up looks good however - I wonder if it would be possible to merge those in and see what the actual split work looks like and what that would look like.  Would it be useful to other people?  It's not too clear to me what use it would be for and your initial description of why you need this isn't shedding much light on things.

lierdakil:
Look at it another way: `AutoLanguageClient` is a messy mix of atom-ide parts and lsp parts, and atom-ide parts are tangled with `ServerManager` even. Doesn't it seem all around better to have those separate, at least from the responsibility separation angle?

I could try to split this into two parts, but honestly, I'm not entirely sure it's worth it since most of it is either making the type checker happy or untangling atom-ide from lsp. Other changes are minor and somewhat sparse, IIRC.

Can we just treat it as refactor only if I drop `BaseLanguageClient` from default package exports?

PR #209, opened 2018-04-18
Do not use autocomplete cache unless explicit trigger character is set

hansec:
During an autocomplete request if no explicit trigger character is set the cache should not be reused as we cannot tell from the trigger character alone (always "") if the prefix is the same. This manifests as a bug when a user changes the prefix after an autocomplete request has been serviced and the autocomplete results are not updated.


damieng:
Thanks!

PR #210, opened 2018-04-21
Compare actual notification message text to determine duplicates

laughedelic:
Notifications deduplication compares only notification type and _details_, but not the message text. This lead to a problem where these two notifications were treated as equal:

screen shot 2018-04-21 at 02 52 10

and therefore the second one was never shown to the user.

Probably `options.detail` was meant to be used for comparison of the notifications text, but it's always the same: server name + project path, so it would treat as duplicates any messages of the same type from the same project/server.


laughedelic:
Probably a further improvement would be comparing the whole `options` object, instead of only `options.detail`. There could be two messages with the same text and different buttons. I think they shouldn't be treated as duplicates.

damieng:
Thanks for this!

PR #213, opened 2018-05-01
Refactors

lierdakil:
As discussed in #206, these are only those refactors I feel are not too controversial.

Commits I think might be *somewhat* controversial are: 

* https://github.com/atom/atom-languageclient/commit/d5f70903b0dd8674f0a413d93463a9ffecf6341e (
Handle unsupported grammar in DocumentSyncAdapter, not ServerManager) -- plain and simple, `ServerManager` handles something it's not responsible for; moving the relevant code makes things that much simpler.
* https://github.com/atom/atom-languageclient/commit/755172ce62506f1a724054fbd0b315f9fb2e480c (
Untangle adapters from ServerManager) -- again, `ServerManager` doesn't even touch adapters that are grafted onto it. I think a better architecture is to use loose coupling via `WeakMap` (even if it might be a bit slower, there are no performance-critical code paths there)
* https://github.com/atom/atom-languageclient/commit/d63fc9ef5124f5832d42428e85e16ac1235318a7 (Untie atomIde.BusySignalService from ServerManager and DocSyncAdapter) -- again, looser coupling. Everything that's done via `BusySignalService` interface could as well be done via a limited version of its `reportBusyWhile` method; hence, it seems like a saner choice to pass only that around.

Each of these changes is small enough. And at least if we agree that less tightly coupled code is generally "better", each I find to be more or less beneficial.

bff82a7 also adds a simple fallback for busy signal via `logger.info`.

If this seems like too much to review at once, reviewing commits one-by-one is also an option -- I tried to keep those more or less self-contained.

### DIFF
--- a/lib/auto-languageclient.ts
+++ b/lib/auto-languageclient.ts
@@ -52,7 +52,7 @@ export type ConnectionType = 'stdio' | 'socket' | 'ipc';
 // implementing startServerProcess/getGrammarScopes/getLanguageName and
 // getServerName.
 export default class AutoLanguageClient {
-  private _disposable = new CompositeDisposable();
+  private _disposable!: CompositeDisposable;
   private _serverManager!: ServerManager;
   private _consoleDelegate?: atomIde.ConsoleService;
   private _linterDelegate?: linter.IndieDelegate;
@@ -231,6 +231,7 @@ export default class AutoLanguageClient {
 
   // Activate does very little for perf reasons - hooks in via ServerManager for later 'activation'
   public activate(): void {
+    this._disposable = new CompositeDisposable();
     this.name = `${this.getLanguageName()} (${this.getServerName()})`;
     this.logger = this.getLogger();
     this._serverManager = new ServerManager(
--- a/lib/auto-languageclient.ts
+++ b/lib/auto-languageclient.ts
@@ -53,27 +53,27 @@ export type ConnectionType = 'stdio' | 'socket' | 'ipc';
 // getServerName.
 export default class AutoLanguageClient {
   private _disposable = new CompositeDisposable();
-  private _serverManager: ServerManager;
-  private _consoleDelegate: atomIde.ConsoleService;
-  private _linterDelegate: linter.IndieDelegate;
-  private _signatureHelpRegistry: atomIde.SignatureHelpRegistry | null;
-  private _lastAutocompleteRequest: AutocompleteRequest;
-  private _isDeactivating: boolean;
+  private _serverManager!: ServerManager;
+  private _consoleDelegate?: atomIde.ConsoleService;
+  private _linterDelegate?: linter.IndieDelegate;
+  private _signatureHelpRegistry?: atomIde.SignatureHelpRegistry;
+  private _lastAutocompleteRequest?: AutocompleteRequest;
+  private _isDeactivating: boolean = false;
 
   // Available if consumeBusySignal is setup
-  protected busySignalService: atomIde.BusySignalService;
+  protected busySignalService?: atomIde.BusySignalService;
 
   protected processStdErr: string = '';
-  protected logger: Logger;
-  protected name: string;
-  protected socket: Socket;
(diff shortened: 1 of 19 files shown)


lib/auto-languageclient.ts
```diff
@@ -47,33 +47,40 @@ import {
...
 // getServerName.
 export default class AutoLanguageClient {
-  private _disposable = new CompositeDisposable();
-  private _serverManager: ServerManager;
-  private _consoleDelegate: atomIde.ConsoleService;
-  private _linterDelegate: linter.IndieDelegate;
-  private _signatureHelpRegistry: atomIde.SignatureHelpRegistry | null;
-  private _lastAutocompleteRequest: AutocompleteRequest;
-  private _isDeactivating: boolean;
+  private _disposable!: CompositeDisposable;
```
lierdakil: Lest I forget, this is actually a bug. If `_disposable` is created on `AutoLanguageClient` construction, and dismissed in `deactivate()`, an activation after deactivation will error due to `add`ing to a disposed `CompositeDisposable`. Hence `_disposable` has to be created in `activate()`.

lierdakil:
Ping.

lib/auto-languageclient.ts
```diff
@@ -687,15 +697,41 @@ export default class AutoLanguageClient {
    * @param filePath path of a file that has changed in the project path
    * @return false => message will not be sent to the language server
    */
-  protected filterChangeWatchedFiles(filePath: string): boolean {
+  protected filterChangeWatchedFiles(_filePath: string): boolean {
```
damieng: What's with all the _ naming on function parameter names?

lierdakil: Since I enabled a couple optional diagnostics [here](https://github.com/atom/atom-languageclient/pull/213/files#diff-e5e546dd2eb0351f813d63d1b39dbc48R14), in particular, `noUnusedParameters`, otherwise TS compiler would complain. If unused parameters' name starts with an underscore, this diagnostic ignores it.

For when this could be useful, consider commit https://github.com/atom/atom-languageclient/pull/213/commits/e6856cbeddab81b3f1a66a001c4d719dae46364c -- `projectPath` is unused there, and the method is private. Hence, either the method should be `protected` to allow overriding, or the `projectPath` parameter is completely pointless and should be removed (I opted for the former). I would be much harder to catch this without `noUnusedParameters`.

damieng: Is there any way of setting this to ignore on protected methods?  It looks like a leaky abstraction to see methods that you can override prefixing parameters with an underscore.

lierdakil: Besides `//@ts-ignore`, which will turn off *all* diagnostics, no, not at the moment, as far as I know.

I can agree that it's leaking a bit of implementation detail, but I personally don't think this is a big deal. For one, overrides are free to use their own name for any variable (or not use any variables at all).

Perhaps, a better solution would be to have `@unused` JSDoc-like decorator, but as of TS 2.8 there's no such thing, and there are no plans to implement anything like this either.

lierdakil: Also, notice that it's not necessarily clear whether or not a `protected` (or for that matter `public`) method can have unused parameters or not. If it's intended to be overridden, sure. If it's not intended to be overridden (for example, it's an utility method which children classes can use), then unused parameters are an indication of an error or leftover code. With the notable exception of `abstract` methods (which *can't* have a default implementation), it's impossible to tell from the code alone whether a given method is intended to be overrideable.

lierdakil:
Ping. Sorry for nagging, but this has been hanging here for a while. Can I do something to expedite the process?

damieng:
I think if we get rid of the _ prefix on unused parameters and switch off that rule we can get this merged.

lierdakil:
Okay, `_` on unused parameters in methods removed and `noUnusedParamters` diagnostic turned off. I left local function parameters prefixed since that doesn't leak anything anywhere.

PR #214, opened 2018-05-01
Clean up atom typings

lierdakil:
This PR is concerned with preferring Atom definitions from DefinitelyTyped, instead of the home-grown variety -- mostly related to autocomplete-plus types.

One thing of note, it replaces `TextEditor.getURI()` with `TextEditor.getPath()`. Several reasons for that: 
-   `getURI` has been semi-deprecated since v0.200 or something like that
-   it's not documented on atom.io
-   it does exactly the same as `getPath`

Not entirely sure how it got there in the first place.


lierdakil:
This is ready to be reviewed/merged.

damieng:
Thanks!

PR #215, opened 2018-05-02
[DO NOT MERGE] Global LS Command Registry to register/unregister and execute commands

### LINKED ISSUE #212: Support `client/registerCapability`
LSP defines `client/registerCapability` request message to register capabilities such as workspace commands on the client (`workspace/executeCommand`).

BoykoAlex:
Fixes #212 
Introduces global LS commands registry where commands are registered. Registry is shared between all LS clients and can be accessed to execute commands.
`AtomEnvironment` has `CommandRegistry` but it's unclear how to utilize it for LS client purposes... Seems like commands in Atom `CommandRegistry` are mainly for the UI (i.e. click or pick command in the command palette) since they don't take parameters...

Don't think this is the final solution for the problem, but maybe it's a starting point to gather some feedback and turn this into a shape of a PR that can be merged.

### DIFF
--- a/lib/adapters/command-adapter.ts
+++ b/lib/adapters/command-adapter.ts
@@ -1,0 +1,116 @@
+import {LanguageClientConnection} from '../languageclient';
+import {ServerCapabilities} from 'vscode-languageserver-protocol';
+import {DisposableLike} from 'atom';
+import * as UUID from 'uuid/v4';
+
+const GLOBAL: any = global;
+
+export class CommandAdapter implements DisposableLike {
+
+    private registrations: Map<string, string[]>;
+
+    constructor(private connection: LanguageClientConnection) {
+        this.registrations = new Map();
+        connection.onRegisterCommand(registration => {
+            if (registration.registerOptions && Array.isArray(registration.registerOptions.commands)) {
+                this.registerCommands(registration.id, registration.registerOptions.commands);
+            }
+        });
+        connection.onUnregisterCommand(unregisteration => this.unregisterCommands(unregisteration.id));
+    }
+
+    initialize(capabilities: ServerCapabilities) {
+        if (capabilities.executeCommandProvider && Array.isArray(capabilities.executeCommandProvider.commands)) {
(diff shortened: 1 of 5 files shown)


lib/adapters/command-adapter.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    dispose() {
+        const cmdRegistry = this.getLspCommandRegistry();
+        this.registrations.forEach(commands => commands.forEach(command => cmdRegistry.unregister(command)));
+        this.registrations.clear();
+    }
+
+    private getLspCommandRegistry(): LspCommandRegistry {
+        if (!GLOBAL.lspCommandRegistry) {
```
daviwil: Shouldn't this be per-server?  Is there an advantage to making it available globally to all instances of atom-languageclient?

BoykoAlex: There is an advantage to it. If this registry is global then it is possible to make language servers communicate between each other. The communication to JDT LS is great for the case of Spring Tools because it's mostly an add-on over the Java tooling. In formation spring tools may want from JDT LS is classpath, search, javadoc etc. 

Example
I have extended LSP with `spring/javadoc` request message with java artifact binding key and project uri as parameters for the message.
The message is received on my Atom language client extension.
Access global command registry and look for command with id `jdt/javadoc` for example which would the command registered by JDT LS (`ide-java` package)
Execute the command and pass the binding key and project URI as parameters to the `jdt/javadoc` command
The command execution goes to the JDT LS and comes back with the result via LSP messages
The javadoc content is extracted from the result of the command execution and sent as a reply to the initial `spring/javadoc` command
Thus, I have javadoc in my spring tools LS from JDT LS at the end of this

There is a related PR for the `ide-java` package that uses the global registry: https://github.com/atom/ide-java/pull/79

Nevertheless, I understand your concern with the global registry... It also introduces new API to work with the registry... I wish there was something in Atom core that could be utilized for this purpose.

PR #236, opened 2018-11-01
Adds outline icon for Struct and EnumMember symbol kind

isundaylee:
`Struct` is practically a `Class`, so I think using `type-class` would be fine. As of `EnumMember`, I think `type-constant` would make sense? But also open to other suggestions.

### DIFF
--- a/lib/adapters/outline-view-adapter.ts
+++ b/lib/adapters/outline-view-adapter.ts
@@ -214,6 +214,10 @@ export default class OutlineViewAdapter {
         return 'type-string';
       case SymbolKind.Variable:
         return 'type-variable';
+      case SymbolKind.Struct:
+        return 'type-class';
+      case SymbolKind.EnumMember:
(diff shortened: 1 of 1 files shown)


damieng:
Struct can map to class for now but I think you can just return 'enum' for enum as it looks like it is supported in the outline view at https://github.com/facebook-atom/atom-ide-ui/blob/master/modules/atom-ide-ui/pkg/atom-ide-outline-view/lib/types.js

isundaylee:
@damieng Just to clarify, by `EnumMember`, I mean the individual constant values defined in an enum (e.g. the `A`, `B`, and `C` in the code snippet below):

```cpp
enum class RandomEnum {
   A = 0,
   B,
   C
};
```

Are you saying that the individual entries should also be mapped to `enum`?

damieng:
Oh, I see.  Yeah, let's stick with constant for that.

PR #245, opened 2018-11-28
Fix logging array wrappers

alexheretic:
Always worth remembering that even after all the typescript type safety, forgetting to write 3 dots can still screw stuff up :stuck_out_tongue_winking_eye:.

### Before
before


damieng:
My mistake, thanks for catching it.

PR #249, opened 2018-12-03
Handling custom requests

laughedelic:
- added missing `onCustomRequest` method
- renamed `onCustom` to `onCustomNotification`

### DIFF
--- a/lib/languageclient.ts
+++ b/lib/languageclient.ts
@@ -22,6 +22,7 @@ export interface KnownRequests {
   [lsp.ShowMessageRequestParams, lsp.MessageActionItem | null];
   'workspace/applyEdit':
   [lsp.ApplyWorkspaceEditParams, lsp.ApplyWorkspaceEditResponse];
+  [custom: string]: [object, object | null];
 }
 
 export type RequestCallback<T extends keyof KnownRequests> =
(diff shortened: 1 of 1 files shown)


lib/languageclient.ts
```diff
@@ -93,15 +94,29 @@ export class LanguageClientConnection extends EventEmitter {
...
   //              The payload from the message is passed to the function.
-  public onCustom(method: string, callback: (obj: object) => void): void {
+  public onCustomNotification(method: string, callback: (obj: object) => void): void {
     this._onNotification({ method }, callback);
   }
 
+  // @deprecated Use onCustomNotification method instead
+  public onCustom(method: string, callback: (obj: object) => void): void {
+    this.onCustomNotification(method, callback);
+  }
```
laughedelic: I'm not sure this is the right way to deprecate in TypeScript and probably it's too harsh to deprecate it at all. I just think `onCustom` name is confusing, because it's only about notifications.

lib/languageclient.ts
```diff
@@ -93,15 +94,29 @@ export class LanguageClientConnection extends EventEmitter {
...
+
+  // Public: Register a callback for a custom request.
+  //
+  // * `method`   A string containing the name of the message to listen for.
+  // * `callback` The function to be called when the message is received.
+  //              The payload from the message is passed to the function.
+  public onCustomRequest(method: string, callback: (obj: object) => Promise<object>): void {
```
laughedelic: Is it ok or the type should be `callback: (obj: object) => Promise<object | null>`?

laughedelic:
Ping.

Could you merge this or tell me if I need to change anything?

laughedelic:
I updated the branch. Can this be reviewed by the new maintainers?

laughedelic:
Moved to https://github.com/atom-community/atom-languageclient/pull/129

PR #250, opened 2018-12-21
support non project root language servers

### LINKED ISSUE #233: project root != language server root
Let's imagine several projects combined in one large git repository. Each of these projects has their own root folder in this repo. They are all related to each other so it makes sense to open them in 1 atom instance instead of opening several atom instances and switching between them.

AFAIK this is currently not supported by the `AutoLanguageClient` because the server always takes the atom root folder as `projectPath` that is then passed to the `startServerProcess` function.

There is already the `determineProjectPath` function on the `ServerManager`:

https://github.com/atom/atom-languageclient/blob/606faa36216657cfd8e1cea9cd2b7974e3a357ce/lib/server-manager.ts#L257-L263

If this can be overwritten by the `AutoLanguageClient` in some way then I think the above mentioned problem is fixed.

What do you think? I think I can prepare a PR for this.

/cc @damieng

lloiser:
fixes #233 

The only thing missing (IMO) was that language packages had to override  `determineProjectPath` so that they can provide an alternate language server root instead of the atom project root folder.

/cc @damieng

### DIFF
--- a/lib/adapters/apply-edit-adapter.ts
+++ b/lib/adapters/apply-edit-adapter.ts
@@ -4,6 +4,7 @@ import {
   LanguageClientConnection,
   ApplyWorkspaceEditParams,
   ApplyWorkspaceEditResponse,
+  TextDocumentEdit,
 } from '../languageclient';
 import {
   TextBuffer,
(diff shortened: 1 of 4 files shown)


damieng:
Sorry I am no longer involved with GitHub or Atom.

lloiser:
Hm.. do you know who might be responsible for this repo now?
sry if I'm bothering you...

akonwi:
I assume @daviwil is still involved and might be able to help identify the new maintainer(s).

lloiser:
FYI: I have published my own version of atom-languageclient to npm
-> https://www.npmjs.com/package/@lloiser/atom-languageclient

Issue #246, opened 2018-12-01
TypeError: Cannot read property 'isIncomplete' of null


Aerijo:
Noticed this package spewing errors into the console

```
/Users/benjamingray/.atom/packages/ide-typescript/node_modules/atom-languageclient/build/lib/adapters/autocomplete-adapter.js:89 Uncaught (in promise) TypeError: Cannot read property 'isIncomplete' of null
    at AutocompleteAdapter.<anonymous> (/Users/benjamingray/.atom/packages/ide-typescript/node_modules/atom-languageclient/build/lib/adapters/autocomplete-adapter.js:89:74)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/benjamingray/.atom/packages/ide-typescript/node_modules/atom-languageclient/build/lib/adapters/autocomplete-adapter.js:4:58)
    at <anonymous>
```

Looks like it's not handling the case where [completions is null](https://github.com/atom/atom-languageclient/blob/27ba74f8c0ea9fb486792676ae7911a37d446bb2/lib/adapters/autocomplete-adapter.ts#L137) before checking a property.

PR #251, opened 2019-01-17
Support null completions

Aerijo:
Adds on to the commit closing #246. The first solution didn't reach far enough, and the possibility of `completions` being null wans't reflected in the types.

This change should comprehensively support a null value.

(Also see [the response section](https://microsoft.github.io/language-server-protocol/specification#textDocument_completion) of the protocol to confirm `null` is a valid return value)


Aerijo:
I've been using this on mine for a while, should work fine

PR #262, opened 2019-03-05
Convert to custom Suggestion type & filter text with `filterText` key

Aerijo:
Converts the internal uses of `ac.AnySuggestion` to the new `Suggestion`, which is an extended version of the original. In particular, it has the property `filterText`, used for custom filtering. It also makes it easier to add features, such as the replacement prefix tracking in #254

Regarding filtering using `filterText`, it seems not even VS Code had a solid idea on the difference between `filterText` and `sortText` for a time
- https://github.com/Microsoft/vscode/issues/26944#issuecomment-303319615
- https://github.com/Microsoft/vscode/issues/26096#issuecomment-306409809

The latest I could find is in that last issue, where they say [paraphrasing] "use sort text for no prefix" and "use filterText for when there's a prefix" (and they base on other things like match index). The [source code](https://github.com/Microsoft/vscode/blob/73ff13f5debf91fc1d7a89a2f336543b2baa1ee5/src/vs/editor/contrib/suggest/completionModel.ts#L219) also supports using `filterText` to sort as well when there's a prefix.

### DIFF
--- a/lib/adapters/autocomplete-adapter.ts
+++ b/lib/adapters/autocomplete-adapter.ts
@@ -20,12 +20,13 @@ import {
   TextEditor,
 } from 'atom';
 import * as ac from 'atom/autocomplete-plus';
+import { Suggestion, TextSuggestion, SnippetSuggestion } from 'atom-ide';
 
 interface SuggestionCacheEntry {
   isIncomplete: boolean;
   triggerPoint: Point;
   triggerChar: string;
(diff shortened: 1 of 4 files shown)


Aerijo:
@50Wliu The `package-lock.json` looks quite different to the one you committed earlier, specifically the version range symbols have been removed. Is this fine?

50Wliu:
Yeah, that's fine. Just commit whatever the result of `npm install` is.

PR #263, opened 2019-03-06
Convert doc comments to TSDoc syntax

Aerijo:
The current comments are very informative, but were not being picked up by `ide-typescript`. This converts them to a more standardised format, which is recognised. 

Note it is not exactly a TSDoc match (which is quite difficult, considering TSDoc does not actually provide a spec yet...). However, it is clearly picked up and parsed by `ide-typescript`, so will suffice.

The only concern would be the `{reference}` syntax, which Atom likes to pick up as something (but not what it's meant to be) but `ide-typescript` ignores. It currently makes a lot of these comments very red.

I plan to merge anyway, short of any suggestions from @damieng or others about why the original syntax was used, because useful doc comments are more valuable than locally broken syntax highlighting IMO.

### DIFF
--- a/lib/adapters/apply-edit-adapter.ts
+++ b/lib/adapters/apply-edit-adapter.ts
@@ -10,9 +10,9 @@ import {
   TextEditor,
 } from 'atom';
 
-/**  Public: Adapts workspace/applyEdit commands to editors. */
+/** Public: Adapts workspace/applyEdit commands to editors. */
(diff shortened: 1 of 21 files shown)


damieng:
The original syntax was based on some syntax elsewhere in Atom but if you can get decent docs out of it or make TypeScript pick it up with alternative syntax go for it!

Issue #183, opened 2018-02-26
Support for custom invocation of code actions' Commands


deankevorkian:
Currently, the default client's flow of using code action works as follows -
1. Client requests server for available code actions according to file & location
2. Client receives the actions and shows them (and every action is associated with a `Command` object)
3. Client chooses one of the code actions in the UI
4. Client calls `workspace/executeCommand` for the Command associated with the code action.

IMO the LSP specs state pretty clearly [here](https://microsoft.github.io/language-server-protocol/specification#textDocument_codeAction) that the response for code actions should return an array of `Command`s that the Server is capable of running - the client should contact the server again with the associated `Command` to executeCommand.

However, not all servers work that way - for example eclipses' java server returns `Command`s it expects the calling client to understand, analyze and act to. (So obviously, non-standard/custom commands).

Since in our default implementation, in step 4 mentioned above, **If I understand correctly**, executeCommand is called automatically for `Command`s the server can't handle,' there's a need to override that behavior (which includes copying redundant code to packages, like `ide-java`, to implement a custom behavior).

Can the atom language client provide some hooks to registering custom commands (some `Map` object that maps a `Command` to a `callback` function can do) for ide-* packages? the way I see it, the client can expose functions to register a custom `Command` callback (without exposing the Map itself) and everytime a `CodeAction` is executed, it will first check if the `Map` contains custom implementation, and if it doesn't - it'll call `workspace/executeCommand` on the server.

thoughts?

deankevorkian:
@damieng could you please take a look at this only to tell if that is something you'd want? if so i'll submit a pr regarding this.

Arcanemagus:
This sounds like a bug in Eclipses Java server. As you already state the spec clearly says that the server should respond to [`textDocument/codeAction`](https://microsoft.github.io/language-server-protocol/specification#textDocument_codeAction) with a list of [`Command`](https://microsoft.github.io/language-server-protocol/specification#command) literals that the server understands.

If Eclipse's Java server is responding with `Command`s that it doesn't understand how to handle, that's a pretty big bug on the part of the server, not something that should be worked around in the clients, after all that's what a standard is for 😉.

deankevorkian:
That is correct, but only correct to this point of time. Back on v2 of the LSP specs it wasn't as clear and explicit as it is now on the recent specs.
[...]

PR #268, opened 2019-05-28
HACK: Support eclipe.jdt.ls's break with the LSP spec

50Wliu:
This is a rather hacky hack to support eclipse.jdt.ls's custom `java.apply.workspaceEdit` command that they expect the _client_ to handle.

Originally I tried to implement a way for clients to specify which commands they wanted to handle as suggested in #183, but I ran into issues with being unable to actually follow through with the command. I can get the client to receive the command fine, but since all the logic is still in atom-languageclient, I can't hook into `applyWorkspaceEdit` or anything of the sort 😓. Maybe I'm missing something that would make this approach feasible?

### DIFF
--- a/lib/adapters/code-action-adapter.ts
+++ b/lib/adapters/code-action-adapter.ts
@@ -89,10 +89,20 @@ export default class CodeActionAdapter {
     connection: LanguageClientConnection,
   ): Promise<void> {
     if (Command.is(command)) {
-      await connection.executeCommand({
-        command: command.command,
-        arguments: command.arguments,
-      });
+      // eclipse.jdt.ls breaks with the spec and expects the client to handle the command
+      // https://github.com/eclipse/eclipse.jdt.ls/issues/376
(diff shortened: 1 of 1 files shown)


Aerijo:
Could it be possible to add a command interface that `ide-java` can hook into on preinitialisation? Like [this method](https://github.com/atom/atom-languageclient/blob/485bb9d706b422456640c9070eee456ef2cf09c0/lib/languageclient.ts#L99-L108) but for commands.

As an aside, it seems they make decisions because "it works in VS Code", not because it conforms with the LSP spec. As such, I will now be referring to the Eclipse language server as a VS Code extension that happens to match some of the LSP. I was only half joking just then.

50Wliu:
Sorry for the late reply. Yeah, I can add a hook, but the issue still is lack of context. I'll need at the minimum a way to apply a workspace edit from within ide-java.

hoihoi500:
Alles komt goed

Issue #13, opened 2017-03-31
textDocument/rename


damieng:
The [textDocument/rename](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#textDocument_rename) protocol lets you rename a symbol and automatically update all associated symbols.

We could implement this with a command and a right-menu button but it would also need a ui for the new symbol unless we could do it inline.

hansonw:
We've got a UI for that in Nuclide as well. It's a tad heavyweight (since it's designed to support more complex refactorings) so I'm not sure if you'd want to start from there or start afresh.

https://github.com/facebook/nuclide/blob/master/pkg/nuclide-refactorizer

The basic flow is right click a symbol, click Refactor in the context menu, select rename, and then type in a new name. There's also an explicit command for quickly just renaming.

lierdakil:
What's the status on this? Minimal implementation should be doable in less than 100 lines... For instance, https://github.com/atom-haskell/hie-lsp-adapter/blob/master/lib/main.js#L20

arxanas:
See also https://github.com/facebook-atom/atom-ide-ui/issues/145#issuecomment-412589283

BenjaminNavarro:
I'm not very familiar with Atom extensions development but I'd like to contribute to this issue. Any tips on where to start?

PR #270, opened 2019-07-02
Add support for rename refactors

lexi-lambda:
This pull request implements rename refactoring support via `textDocument/rename`, resolving #13. It is compatible with the protocol expected by the most recent release of `atom-ide-ui`, which is `0.13.0`. Here’s a demonstration using the `ide-haskell-hie` backend:

(It’s not the fastest backend in the world, but it gets the job done eventually.)

Awkwardly, this implementation *doesn’t* work with the latest version of `atom-ide-ui` in [the archived source code](https://github.com/facebookarchive/atom-ide-ui), since it was changed just after the final release to use a slightly different protocol. I don’t think that’s a big deal, though, since it’s always possible to bump the service version if those changes ever make it into something released.

### DIFF
--- a/lib/adapters/rename-adapter.ts
+++ b/lib/adapters/rename-adapter.ts
@@ -1,0 +1,75 @@
+import * as atomIde from 'atom-ide';
+import Convert from '../convert';
+import {
+  Point,
+  TextEditor,
+} from 'atom';
+import {
+  LanguageClientConnection,
+  RenameParams,
+  ServerCapabilities,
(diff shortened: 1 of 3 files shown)


Aerijo:
Works for me on the last `atom-ide-ui` release. Do you know if the service has a specification anywhere? Especially as you mention they were working on a revised version.

lexi-lambda:
> Do you know if the service has a specification anywhere?

I have no idea, but if it does, I don’t know about it.

Issue #166, opened 2018-01-07
npm ERR! Failed at the atom-languageclient@0.1.2 postinstall script 'postinstall-build --only-as-dependency build "npm run compile"'.


Johnathon332:
I am trying to install sonarlint for atom text editor but have come across this error.

```
Installing “sonarlint@1.0.2” failed.Hide output…

> atom-languageclient@0.1.2 postinstall C:\Users\JOHNAT~1\AppData\Local\Temp\apm-install-dir-11806-10740-1p6l9uz.gd9xczyqfr\node_modules\sonarlint\node_modules\atom-languageclient
> postinstall-build --only-as-dependency build "npm run compile"


> atom-languageclient@0.1.2 postinstall C:\Users\JOHNAT~1\AppData\Local\Temp\apm-install-dir-11806-10740-1p6l9uz.gd9xczyqfr\node_modules\sonarlint\node_modules\atom-languageclient
> postinstall-build --only-as-dependency build "npm run compile"


> atom-languageclient@0.1.2 prepublish C:\Users\JOHNAT~1\AppData\Local\Temp\apm-install-dir-11806-10740-1p6l9uz.gd9xczyqfr\node_modules\sonarlint\node_modules\atom-languageclient
> npm run clean && npm run compile


> atom-languageclient@0.1.2 clean C:\Users\JOHNAT~1\AppData\Local\Temp\apm-install-dir-11806-10740-1p6l9uz.gd9xczyqfr\node_modules\sonarlint\node_modules\atom-languageclient
> rm -rf build

C:\Users\JOHNAT~1\AppData\Local\Temp\apm-install-dir-11806-10740-1p6l9uz.gd9xczyqfr
`-- (empty)


npm WARN deprecated tough-cookie@2.2.2: ReDoS vulnerability parsing Set-Cookie https://nodesecurity.io/advisories/130
npm WARN deprecated node-uuid@1.4.8: Use uuid module instead
'rm' is not recognized as an internal or external command,
operable program or batch file.

npm ERR! Windows_NT 10.0.14393
npm ERR! argv "C:\\Users\\Johnathon332\\AppData\\Local\\atom\\app-1.23.2\\resources\\app\\apm\\bin\\node.exe" "C:\\Users\\Johnathon332\\AppData\\Local\\atom\\app-1.23.2\\resources\\app\\apm\\node_modules\\npm\\bin\\npm-cli.js" "run" "clean"
npm ERR! node v6.9.5
[...]

PR #276, opened 2019-10-01
use rimraf

### LINKED ISSUE #275: Doesn't install on windows
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

### Description

Won't install on windows. 

`'rm' is not recognized as an internal or external command`

`npm run clean` runs `rm -rf build`

related https://github.com/Gert-dev/php-ide-serenata/issues/477

### Steps to Reproduce

1. run `npm i` on windows

**Expected behavior:**

installs

**Actual behavior:**

errors

**Reproduces how often:**

100%

UziTech:
### Identify the Bug

Doesn't install on window

fixes #275 

### Description of the Change

use `rimraf` instead of `rm -rf`

### Possible Drawbacks

none

### Verification Process

installed on windows

### Release Notes

Fix windows install

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
       "requires": {
         "diff": "4.0.1",
         "etch": "0.14.0",
-        "grim": "2.0.2",
+        "grim": "^2.0.1",
         "klaw-sync": "6.0.0",
         "less": "3.9.0",
         "mocha": "6.0.2",
@@ -26,7 +26,7 @@
           "integrity": "sha512-mel+jf7nrtEl5Pn1Qx46zARXKDpBbvzezse7p7LqINmdoIk8PYP5SySaxEmYv6TZ0JyEKA1hsCId6DIhgITtWQ==",
           "dev": true,
           "requires": {
-            "ms": "2.1.1"
+            "ms": "^2.1.1"
           }
         },
         "diff": {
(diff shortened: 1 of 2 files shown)


Gert-dev:
This makes me curious, do other language clients work on Windows at all? It seems highly unlikely that we're the first to encounter this error as there have been many language clients for much longer. 

php-ide-serenata has only just switched and Windows users now immediately encounter this error.

UziTech:
It also seems fairly strange that tests pass on [appveyor](https://ci.appveyor.com/project/Atom/atom-languageclient/branch/master)

Gert-dev:
Hmm, I think I see what the problem is (#166), it is because Serenata is using an unstable version of the atom-languageclient - intentionally, I might add, since I need the targeted commit to fix an issue with autocompletion.

It is likely it'll be solved by targeting a stable release, but autocompletion will be bugged for users. If a new stable release is performed here, that would likely also fix the problem (but this fix might as well).

UziTech:
@Aerijo @50Wliu @damieng @atom Any feedback?

damieng:
I haven't worked on Atom or the language client in well over a year, sorry.

UziTech:
@atom Could we get this merged so it works on windows and a new release?

PR #285, opened 2020-05-12
Support non-boolean server capabilities + update dependencies

alexheretic:
Server capabilities which were boolean are no longer only booleans. This PR adds initial support for non-boolean server capabilities by at least not treating them as `false`.

I've also update the language server types dependencies, typescript version + some of the dev-dependencies.

This fix is also available in my fork:
```json
"dependencies": {
  "atom-languageclient": "github:alexheretic/atom-languageclient#build",
}
```


hoihoi500:
Added a updated nean bootclean server cap, to test my buseniss. GR

PR #286, opened 2020-06-09
mm

hoihoi500:
We must be able to understand the purpose of your change from this description. If we can't get a good idea of the benefits of the change from the description here, the pull request may be closed at the maintainers' discretion.

Only for private only use

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,4 @@
-Copyright (c) 2017 GitHub Inc.
+*Copyright (c) 2017 GitHub Inc.+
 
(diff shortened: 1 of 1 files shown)


hoihoi500:
For testing private

Arcanemagus:
Thanks for reaching out!

We require the template to be filled out on all new issues and pull requests. We do this so that we can be certain we have all the information we need to address your submission efficiently. This allows the maintainers to spend more time fixing bugs, implementing enhancements, and reviewing and merging pull requests.

Thanks for understanding and meeting us halfway :grinning:

